PR #9, opened 2019-03-12
Add an API endpoint to fetch user information

feng-tao:
The user model for "Amundsen People" could be found in https://github.com/lyft/amundsendatabuilder/blob/master/databuilder/models/user.py.

Add an API endpoint to fetch user based on user_id(email).

Test the endpoint locally with staging neo4j data:
```
curl http://localhost:5000/user/'tfeng@lyft.com'
{"email": "tfeng@lyft.com", "first_name": "Tao", "last_name": "Feng", "full_name": "Tao Feng", "is_active": true, "github_username": "", "slack_id": "", "team_name": "Data Platform", "employee_type": "teamMember"}
```

The databuilder has some issues on `github_user_name` field, but that would be a separate pr to fix.
Endpoint for bookmark, own, used three relationships will be in a separate pr.

### DIFF
--- a/metadata_service/proxy/neo4j_proxy.py
+++ b/metadata_service/proxy/neo4j_proxy.py
@@ -16,6 +16,7 @@ from metadata_service.entity.popular_table import PopularTable
 from metadata_service.entity.table_detail import Application, Column, Reader, Source, \
     Statistics, Table, Tag, User, Watermark
 from metadata_service.entity.tag_detail import TagDetail
+from metadata_service.entity.user_detail import User as UserEntity
 from metadata_service.exception import NotFoundException
 from metadata_service.proxy.statsd_utilities import timer_with_counter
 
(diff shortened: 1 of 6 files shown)


feng-tao:
PTAL @jinhyukchang @ttannis @danwom

metadata_service/proxy/neo4j_proxy.py
```diff
@@ -682,6 +683,35 @@ def get_popular_tables(self, *, num_entries: int =10) -> List[PopularTable]:
...
+            result = UserEntity(email=record['email'],
+                                first_name=record.get('first_name'),
+                                last_name=record.get('last_name'),
+                                full_name=record.get('full_name'),
+                                is_active=record.get('is_active'),
+                                github_username=record.get('github_username'),
+                                team_name=record.get('team_name'),
+                                slack_id=record.get('slack_id'),
+                                employee_type=record.get('employee_type'))
+        return result
```
jinhyukchang: I think we can throw `NotFoundException` if not exist instead of returning `None`

feng-tao: updated

PR #10, opened 2019-03-13
Add manager info in user endpoint payload

feng-tao:
Add the manager full name as user endpoint payload.

Note: We need to verify whether we have the data correctly in manager.

### DIFF
--- a/metadata_service/proxy/neo4j_proxy.py
+++ b/metadata_service/proxy/neo4j_proxy.py
@@ -693,7 +693,9 @@ class Neo4jProxy:
         """
 
         query = textwrap.dedent("""
-        MATCH (user:User {key: $user_id}) RETURN user as user_record
+        MATCH (user:User {key: $user_id})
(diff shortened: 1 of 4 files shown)


feng-tao:
cc @ttannis @jinhyukchang @danwom

metadata_service/proxy/neo4j_proxy.py
```diff
@@ -693,15 +693,19 @@ def get_user_detail(self, *, user_id: str) -> Union[UserEntity, None]:
...
 
         record = self._execute_cypher_query(statement=query,
                                             param_dict={'user_id': user_id})
         if not record:
             raise NotFoundException('User {user_id} '
                                     'not found in the graph'.format(user_id=user_id))
-        record = record.single()['user_record']
+        single_result = record.single()
+        record = single_result['user_record']
+        manager_record = single_result['manager_record']
```
jinhyukchang: Could this cause NPE?

feng-tao: updated

PR #11, opened 2019-03-14
[Amundsen People] Add API to build relationship between user and table

feng-tao:
Test locally with staging data which works:

action in { own, follow, read}
```
curl http://localhost:5000/user/tfeng@lyft.com/action/
curl -v -x PUT http://localhost:5000/user/tfeng@lyft.com/action/hive%3A%2F%2Fgold.default%2Ftable_name
curl -v -x DELETE http://localhost:5000/user/tfeng@lyft.com/action/hive%3A%2F%2Fgold.default%2Ftable_name
```

todo:
1. will create a separate pr to refactor existing table_owner endpoint to use the new generic endpoint.
2. currently it doesn't take type as url input parameter which requires larger refactor. I plan to do(create a ticket) once we have the dashboard feature coming in.

Design qq:
1. do we need to support put/delete for read or user API?
2. do we need to allow FE to pass a threshold to the read API endpoint? Currently based on staging data, I read / use 28 tables which I am not sure if we want to display all of them.

### DIFF
--- a/metadata_service/__init__.py
+++ b/metadata_service/__init__.py
@@ -15,7 +15,7 @@ from metadata_service.api.system import Neo4jDetailAPI
 from metadata_service.api.table \
     import TableDetailAPI, TableOwnerAPI, TableTagAPI, TableDescriptionAPI
 from metadata_service.api.tag import TagAPI
-from metadata_service.api.user import UserDetailAPI
+from metadata_service.api.user import UserDetailAPI, UserBookmarkAPI, UserOwnAPI, UserReadAPI
 
 # For customized flask use below arguments to override.
 FLASK_APP_MODULE_NAME = os.getenv('FLASK_APP_MODULE_NAME')
@@ -86,7 +86,15 @@ def create_app(*, config_module_class: str) -> Flask:
                      '/tags/')
     api.add_resource(UserDetailAPI,
                      '/user/<path:user_id>')
-
+    api.add_resource(UserBookmarkAPI,
+                     '/user/<path:user_id>/bookmark/',
+                     '/user/<path:user_id>/bookmark/<path:table_uri>')
+    api.add_resource(UserOwnAPI,
+                     '/user/<path:user_id>/own/',
+                     '/user/<path:user_id>/own/<path:table_uri>')
+    api.add_resource(UserReadAPI,
+                     '/user/<path:user_id>/read/',
+                     '/user/<path:user_id>/read/<path:table_uri>')
     app.register_blueprint(api_bp)
 
     return app
--- a/metadata_service/__init__.py
+++ b/metadata_service/__init__.py
@@ -15,7 +15,7 @@ from metadata_service.api.system import Neo4jDetailAPI
 from metadata_service.api.table \
     import TableDetailAPI, TableOwnerAPI, TableTagAPI, TableDescriptionAPI
 from metadata_service.api.tag import TagAPI
-from metadata_service.api.user import UserDetailAPI, UserBookmarkAPI, UserOwnAPI, UserReadAPI
+from metadata_service.api.user import UserDetailAPI, UserFollowAPI, UserOwnAPI, UserReadAPI
 
 # For customized flask use below arguments to override.
 FLASK_APP_MODULE_NAME = os.getenv('FLASK_APP_MODULE_NAME')
@@ -86,15 +86,15 @@ def create_app(*, config_module_class: str) -> Flask:
                      '/tags/')
     api.add_resource(UserDetailAPI,
                      '/user/<path:user_id>')
-    api.add_resource(UserBookmarkAPI,
-                     '/user/<path:user_id>/bookmark/',
-                     '/user/<path:user_id>/bookmark/<path:table_uri>')
+    api.add_resource(UserFollowAPI,
+                     '/user/<path:user_id>/follow/',
+                     '/user/<path:user_id>/follow/<resource_type>/<path:table_uri>')
     api.add_resource(UserOwnAPI,
                      '/user/<path:user_id>/own/',
-                     '/user/<path:user_id>/own/<path:table_uri>')
+                     '/user/<path:user_id>/own/<resource_type>/<path:table_uri>')
     api.add_resource(UserReadAPI,
                      '/user/<path:user_id>/read/',
-                     '/user/<path:user_id>/read/<path:table_uri>')
+                     '/user/<path:user_id>/read/<resource_type>/<path:table_uri>')
     app.register_blueprint(api_bp)
 
     return app
--- a/metadata_service/api/user.py
+++ b/metadata_service/api/user.py
@@ -1,5 +1,5 @@
 from http import HTTPStatus
-from typing import Iterable, Mapping, Union
+from typing import Iterable, Mapping, Union, Optional, Any, Tuple
 
 from flask_restful import Resource, fields, marshal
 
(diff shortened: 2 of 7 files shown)


feng-tao:
cc @danwom @ttannis @jinhyukchang

metadata_service/proxy/neo4j_proxy.py
```diff
@@ -715,9 +719,136 @@ def get_user_detail(self, *, user_id: str) -> Union[UserEntity, None]:
...
+        :param reverse_relation:
+        :return:
+        """
+        upsert_user_query = textwrap.dedent("""
+        MERGE (u:User {key: $user_email})
+        on CREATE SET u={email: $user_email, key: $user_email}
+        on MATCH SET u={email: $user_email, key: $user_email}
+        """)
+
+        user_email = 'key: "{user_email}"'.format(user_email=user)
```
feng-tao: the reason to do this is because it seems neo4j python client doesn't support relation parameterized.  Hence I need to construct the actual query url before sending over to the actual driver based on different relation ship. I intent to make this function generic as we will have multiple relationship between user and resources(Table, Dashboard, Kafka  Streams etc)

metadata_service/__init__.py
```diff
@@ -86,7 +86,15 @@ def create_app(*, config_module_class: str) -> Flask:
                      '/tags/')
     api.add_resource(UserDetailAPI,
                      '/user/<path:user_id>')
-
+    api.add_resource(UserBookmarkAPI,
+                     '/user/<path:user_id>/bookmark/',
+                     '/user/<path:user_id>/bookmark/<path:table_uri>')
```
jinhyukchang: For starter, how about adding type here?
`/user/<path:user_id>/bookmark/<type>/<path:table_uri>`
I think we can simplify the implementation by checking type is always table for now.

metadata_service/api/user.py
```diff
@@ -31,8 +36,211 @@ def __init__(self) -> None:
...
+class UserBookmarkAPI(Resource):
+    """
+    Build get / put API to support user bookmark resource features.
+    It will create a relationship(follow / followed_by) between user and resources(table, dashboard etc)
+    """
+
+    def __init__(self) -> None:
+        self.neo4j = neo4j_proxy.get_neo4j()
+
+    def get(self, user_id: str) -> Tuple:
```
jinhyukchang: s/Tuple/Iterable ?

metadata_service/api/user.py
```diff
@@ -31,8 +36,211 @@ def __init__(self) -> None:
...
+            return marshal(user, user_detail_fields), HTTPStatus.OK
 
         except NotFoundException:
             return {'message': 'User id {} does not exist'.format(user_id)}, HTTPStatus.NOT_FOUND
+
+
+class UserBookmarkAPI(Resource):
+    """
+    Build get / put API to support user bookmark resource features.
+    It will create a relationship(follow / followed_by) between user and resources(table, dashboard etc)
```
jinhyukchang: I wonder the difference between bookmark and follow. If it's the same feature, how about unifying the term?

metadata_service/api/user.py
```diff
@@ -31,8 +36,211 @@ def __init__(self) -> None:
...
+    def get(self, user_id: str) -> Tuple:
+        """
+        Return a list of resources that user has followed
+
+        :param user_id:
+        :return:
+        """
+        try:
+            resources = self.neo4j.get_resources_by_user_relation(user_id=user_id,
+                                                                  relation='FOLLOW')
```
jinhyukchang: I think we are not ready for abstraction at this moment. I'd rather explicitly call `get_table_by_user_relation` until we design the abstraction especially because it's returning table detail payload. Thought?

Also, can we have relation as enum? It seems it will be increasing and good to keep it in one place as well as it can also hide information of bi-direction relation value from this method.

metadata_service/api/user.py
```diff
@@ -31,8 +36,211 @@ def __init__(self) -> None:
...
+
+        except NotFoundException:
+            return {'message': 'user_id {} does not exist'.format(user_id)}, HTTPStatus.NOT_FOUND
+
+        except Exception:
+            return {'message': 'Internal server error!'}, HTTPStatus.INTERNAL_SERVER_ERROR
+
+    def put(self, user_id: str, table_uri: str) -> Iterable[Union[Mapping, int, None]]:
+        """
+        Create the relationship between user and resources.
```
jinhyukchang: Nit: "Create the relationship" --> "Create the follow relationship" ?

metadata_service/api/user.py
```diff
@@ -31,8 +36,211 @@ def __init__(self) -> None:
...
+        todo: It will need to refactor all neo4j proxy api to take a type argument.
+
+        :param user_id:
+        :param table_uri:
+        :return:
+        """
+        try:
+            self.neo4j.add_resource_relation_by_user(table_uri=table_uri,
```
jinhyukchang: Same comment here. (Explicit on table, and enum relation)

metadata_service/api/user.py
```diff
@@ -31,8 +36,211 @@ def __init__(self) -> None:
...
+                                                              table_uri)}, HTTPStatus.OK
+        except Exception as e:
+            return {'message': 'The user {} for table_uri {} '
+                               'is not added successfully'.format(user_id,
+                                                                  table_uri)}, \
+                HTTPStatus.INTERNAL_SERVER_ERROR
+
+    def delete(self, user_id: str, table_uri: str) -> Iterable[Union[Mapping, int, None]]:
+        """
+        Delete the relationship between user and resources.
```
jinhyukchang: Nit: "Delete the relationship" --> "Delete the follow relationship"

metadata_service/api/user.py
```diff
@@ -31,8 +36,211 @@ def __init__(self) -> None:
...
+                                                                table_uri)}, HTTPStatus.OK
+        except Exception:
+            return {'message': 'The owner {} for table_uri {} '
+                               'is not deleted successfully'.format(user_id,
+                                                                    table_uri)}, HTTPStatus.INTERNAL_SERVER_ERROR
+
+
+class UserReadAPI(Resource):
+    """
+    Build get / put API to support user bookmark resource features.
```
jinhyukchang: s/bookmark/read

metadata_service/api/user.py
```diff
@@ -31,8 +36,211 @@ def __init__(self) -> None:
...
+    It will create a relationship(read / read_by) between user and resources(table, dashboard etc)
+
+    todo: do we need put / delete to this API? And do we need to put a threshold on the get?
```
jinhyukchang: I don't think we have a use case to mutate read record. We can remove it unless we have a use case.

For read, did we define how many tables we will display as well as sorting rule? We may need pagination or truncation

cc: @danwom

metadata_service/proxy/neo4j_proxy.py
```diff
@@ -715,9 +719,136 @@ def get_user_detail(self, *, user_id: str) -> Union[UserEntity, None]:
...
+            raise NotFoundException('User {user_id} does not {relation} '
+                                    'any resources'.format(user_id=user_id,
+                                                           relation=relation))
+        results = []
+        table_records = record.single().get('table_records', [])
+
+        for record in table_records:
+            # todo: decide whether we want to return a list of table entities or just table_uri
+            results.append(self.get_table(table_uri=record['key']))
+        return {'table': results}
```
jinhyukchang: Can we return defined class just as other methods?

metadata_service/proxy/neo4j_proxy.py
```diff
@@ -715,9 +719,136 @@ def get_user_detail(self, *, user_id: str) -> Union[UserEntity, None]:
...
+
+    @timer_with_counter
+    def add_resource_relation_by_user(self, *,
+                                      table_uri: str,
+                                      user: str,
```
jinhyukchang: Nit: s/user/user_email?

metadata_service/proxy/neo4j_proxy.py
```diff
@@ -715,9 +719,136 @@ def get_user_detail(self, *, user_id: str) -> Union[UserEntity, None]:
...
+                raise RuntimeError('Failed to create relation between '
+                                   'user {user} and table {tbl}'.format(user=user,
+                                                                        tbl=table_uri))
+        except Exception as e:
+            if not tx.closed():
+                tx.rollback()
+            # propagate the exception back to api
+            raise e
+        finally:
+            tx.commit()
```
jinhyukchang: This will throw exception if it's rolled back shadowing original exception.
Could you refer to this one? https://github.com/lyft/amundsenmetadatalibrary/blob/master/metadata_service/proxy/neo4j_proxy.py#L319

metadata_service/proxy/neo4j_proxy.py
```diff
@@ -715,9 +719,136 @@ def get_user_detail(self, *, user_id: str) -> Union[UserEntity, None]:
...
+        try:
+            tx = self._driver.session().begin_transaction()
+            tx.run(delete_query, {})
+        except Exception as e:
+            # propagate the exception back to api
+            if not tx.closed():
+                tx.rollback()
+            raise e
+        finally:
+            tx.commit()
```
jinhyukchang: Same comment as above

metadata_service/__init__.py
feng-tao: I change it, but call it resource_type as type is a key_word in python.

metadata_service/api/user.py
feng-tao: correct, make sense, let me remove the two mutate methods.

In terms of sorting, let's keep it simple in first pr(I put this qq in this pr description )

feng-tao:
cc @jinhyukchang 
pr updated

PR #18, opened 2019-03-27
Updates atlas proxy implementation to match terminology

verdan:
In order to make it consistent with the existing terms of Amundsen, this PR updates the way it was using db and schema name to generate the table URI. 
We'll make a more concrete discussion about the terms in an issue opened in frontend library https://github.com/lyft/amundsenfrontendlibrary/issues/61

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,21 +65,4 @@ This way Metadata service will use production config in production environment. 
 - Typing hints: Amundsen Metadata service also utilizes [Typing hint](https://docs.python.org/3/library/typing.html "Typing hint") for better readability.
 
 ## Code structure
-Amundsen metadata service consists of three packages, API, Entity, and Proxy.
-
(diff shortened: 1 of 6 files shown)


verdan:
@feng-tao @jinhyukchang ready for review.

jinhyukchang:
@verdan I believe this PR is contingent to this [issue](https://github.com/lyft/amundsenfrontendlibrary/issues/61). I just replied to the issue, and let me know what you think.

Thanks for the suggestion!
Jin

verdan:
@jinhyukchang this PR is still valid, as previously I was assuming `database` as the database name, `schema` as the database schema like public etc., and was sending the data from atlas proxy based on that. This PR would fix that issue for atlas proxy.

verdan:
@feng-tao nope. I already have the next PR ready, was waiting for this one. And i will bump the version in that one.

feng-tao:
sounds good.

PR #19, opened 2019-04-03
Table Detail APIs improvements

verdan:
This PR extends the functionality of Apache Atlas, and using existing attributes of Atlas, it implements the owner, description, and tags functionality on the detail pages. 
Added Todos and FixMe's where needed for now. 

FYI: @feng-tao @jinhyukchang

### DIFF
--- a/metadata_service/proxy/atlas_proxy.py
+++ b/metadata_service/proxy/atlas_proxy.py
@@ -1,12 +1,13 @@
 import logging
 import re
-from typing import Union, List, Dict, Any
+from typing import Union, List, Dict, Any, Tuple
 
 from atlasclient.client import Atlas
 from atlasclient.exceptions import BadRequest
+from metadata_service.entity.tag_detail import TagDetail
 
 from metadata_service.entity.popular_table import PopularTable
-from metadata_service.entity.table_detail import Table
+from metadata_service.entity.table_detail import Table, User, Tag
 from metadata_service.entity.user_detail import User as UserEntity
 from metadata_service.exception import NotFoundException
 from metadata_service.proxy import BaseProxy
@@ -18,6 +19,7 @@ LOGGER = logging.getLogger(__name__)
 class AtlasProxy(BaseProxy):
     """
     Atlas Proxy client for the amundsen metadata
+    {ATLAS_API_DOCS} = https://atlas.apache.org/api/v2/
     """
     TABLE_ENTITY = 'Table'
     DB_KEY = 'db'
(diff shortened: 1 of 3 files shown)


verdan:
@feng-tao, second thought, I did revert back the version to 1.0.3. I am planning to work on the column information for table detail page next. So it would make more sense to upgrade pypi version at that time.

feng-tao:
@verdan , sounds good, will take a look later today :)

metadata_service/proxy/atlas_proxy.py
```diff
@@ -72,37 +74,61 @@ def _extract_info_from_uri(self, *, table_uri: str) -> Dict:
...
         try:
+            attrs = table_details['attributes']
+            rel_attrs = table_details['relationshipAttributes']
+
+            tags = []
+            for classification in table_details.get("classifications", list()):
+                tags.append(
+                    Tag(
+                        tag_name=classification.get('typeName'),
+                        tag_type=classification.get('typeName')
```
jinhyukchang: Could you use "default" in tag_type? We are currently not using `tag_type`, but filling it with "default".

metadata_service/proxy/atlas_proxy.py
```diff
@@ -72,37 +74,61 @@ def _extract_info_from_uri(self, *, table_uri: str) -> Dict:
...
         """
-        Gathers all the information needed for the Table Detail Page.
-        It tries to get the table information from
+        Fetch information from table_uri and then find the appropriate entity
+        The reason, we're not returning the entity_unique_attribute().entity
+        directly is because the entity_unique_attribute() return entity Object
+        that can be used for update purposes,
+        while entity_unique_attribute().entity only returns the dictionary
         :param table_uri:
         :return:
```
jinhyukchang: Could you add some explanation about the return type's schema to know what data is available? (It's because the reader will only know by looking at the code who consumes the data, but still hard to have full picture)

metadata_service/proxy/atlas_proxy.py
```diff
@@ -116,22 +142,68 @@ def delete_owner(self, *, table_uri: str, owner: str) -> None:
...
+        API Ref: /resource_EntityREST.html#resource_EntityREST_deleteClassification_DELETE
+        :param table_uri:
+        :param tag:
+        :return:
+        """
+        try:
+            entity, _ = self._get_table_entity(table_uri=table_uri)
+            guid_entity = self._driver.entity_guid(entity.entity['guid'])
+            guid_entity.classifications(tag).delete()
+        except Exception as ex:
```
jinhyukchang: This could blind us from real failure. Is there some way to filter out this Exception that happens all the time?

metadata_service/proxy/atlas_proxy.py
```diff
@@ -116,22 +142,68 @@ def delete_owner(self, *, table_uri: str, owner: str) -> None:
         pass
 
     def add_owner(self, *, table_uri: str, owner: str) -> None:
-        pass
+        """
+        It simply replaces the owner field in atlas with the new string.
+        FixMe (Verdan): Implement multiple data owners and
```
jinhyukchang: @verdan Is this limitation of Atlas or specific to ING?

metadata_service/proxy/atlas_proxy.py
verdan: ah, good to know. will update this, thanks 👍

metadata_service/proxy/atlas_proxy.py
verdan: There could be multiple exceptions: 
- Tag Deleted: But there still some tags associated. Can't find this tag exception
- Last Tag Deleted: Different exceptions, sometimes it raises no tags associated, and sometimes it says can't find this tag. 

I will add a FixMe here, and will debug this issue once I'll improve the atlas client library. (planning to fork that one as the maintainer is not responsive)

metadata_service/proxy/atlas_proxy.py
verdan: The owner field type in Atlas is a string by default. I'll have a chat my Bolke and other team members to come up with a better solution.

PR #21, opened 2019-04-09
Implement the column information for atlas proxy

verdan:
This PR implements the column feature for the table detail page on Amundsen. 
Also updates the atlasclient version and introduced the new version for Amundsen metadata library itself.

### DIFF
--- a/metadata_service/proxy/atlas_proxy.py
+++ b/metadata_service/proxy/atlas_proxy.py
@@ -9,7 +9,7 @@ from atlasclient.models import EntityUniqueAttribute
 from metadata_service.entity.tag_detail import TagDetail
 
 from metadata_service.entity.popular_table import PopularTable
-from metadata_service.entity.table_detail import Table, User, Tag
+from metadata_service.entity.table_detail import Table, User, Tag, Column
 from metadata_service.entity.user_detail import User as UserEntity
 from metadata_service.exception import NotFoundException
 from metadata_service.proxy import BaseProxy
@@ -122,6 +122,19 @@ class AtlasProxy(BaseProxy):
                     )
                 )
 
+            columns = []
+            for column in rel_attrs.get('columns', list()):
+                col_entity = self._driver.entity_guid(column['guid'])
+                col_attrs = col_entity.entity['attributes']
+                columns.append(
+                    Column(
+                        name=col_attrs.get(self.NAME_KEY),
+                        description=col_attrs.get('description'),
+                        col_type=col_attrs.get('type'),
+                        sort_order=col_attrs.get('position'),
+                    )
+                )
+
             table = Table(database=table_info['entity'],
                           cluster=table_info['cluster'],
                           schema=table_info['db'],
@@ -129,7 +142,7 @@ class AtlasProxy(BaseProxy):
                           tags=tags,
                           description=attrs.get('description'),
                           owners=[User(email=attrs.get('owner'))],
-                          columns=rel_attrs.get('columns'),
+                          columns=columns,
                           last_updated_timestamp=table_details.get('updateTime'))
 
             return table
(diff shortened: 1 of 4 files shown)


verdan:
@jinhyukchang @feng-tao ready to review.

metadata_service/proxy/atlas_proxy.py
```diff
@@ -93,43 +94,78 @@ def _get_table_entity(self, *, table_uri: str) -> Tuple[EntityUniqueAttribute, D
                 table_info['entity'],
                 qualifiedName=table_info.get('name')), table_info
         except Exception as ex:
-            LOGGER.exception('Table not found. {}'.format(str(ex)))
+            LOGGER.exception(f'Table not found. {str(ex)}')
```
feng-tao: didn't know about that before. just read (https://cito.github.io/blog/f-strings/). pretty neat!

metadata_service/proxy/atlas_proxy.py
```diff
@@ -93,43 +94,78 @@ def _get_table_entity(self, *, table_uri: str) -> Tuple[EntityUniqueAttribute, D
...
         table_details = entity.entity
 
         try:
             attrs = table_details['attributes']
-            rel_attrs = table_details['relationshipAttributes']
 
             tags = []
-            for classification in table_details.get("classifications", list()):
+            # Using or in case, if the key 'classifications' is there with a None
+            for classification in table_details.get("classifications") or list():
```
feng-tao: never use this kinda syntax before, is it new in >= python3.6, is it equivalent to  
```
for classification in table_details.get("classifications", list())
```
?

metadata_service/proxy/atlas_proxy.py
```diff
@@ -93,43 +94,78 @@ def _get_table_entity(self, *, table_uri: str) -> Tuple[EntityUniqueAttribute, D
...
                     Tag(
                         tag_name=classification.get('typeName'),
                         tag_type="default"
                     )
                 )
 
+            columns = []
+            for column in attrs.get('columns') or list():
+                col_entity = entity.referredEntities[column['guid']]
+                col_attrs = col_entity['attributes']
```
feng-tao: does attributes key always exist in col_entity?

metadata_service/proxy/atlas_proxy.py
verdan: yes, f strings are really cool :)

metadata_service/proxy/atlas_proxy.py
verdan: that's a tricky one. 
Let me explain this with example. 

```python
# Suppose you have a dict with some keys with values as 'None'
>>> dummy_dict = {'alpha': 'ALPHA', 'beta': None}

>>> dummy_dict.get('alpha')
'ALPHA'        # Output

>>> dummy_dict.get('beta')
# Output nothing/None

# Access an item which is not in the dictionary, KeyError, then 'default' value of dict.get() would be used.
>>> for item in dummy_dict.get('gamma', list()):
...    print(item)
... 

# Access an item of a dictionary that has the value 'None' then Boom!
>>> for item in dummy_dict.get('beta', list()):
...    print(item)
... 
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: 'NoneType' object is not iterable

# So to avoid the above error, you can do something like this. 
>>> for item in dummy_dict.get('beta') or list():
...     print(item)
...

# The above is same as the following
>>> value = dummy_dict.get('beta') or list()
>>> for item in value:
...     print(item)
...
```

metadata_service/proxy/atlas_proxy.py
verdan: True. Each entity in Apache Atlas has the `attributes` key for sure.

metadata_service/proxy/atlas_proxy.py
feng-tao: got it, thanks.

feng-tao:
lgtm, thanks @verdan

PR #29, opened 2019-04-24
add is_view field to TableDetailAPI payload

youngyjd:
This PR is to add `is_view` field to TableDetailAPI payload, so frontend can get the value and differentiate Views from Tables.

### DIFF
--- a/metadata_service/proxy/neo4j_proxy.py
+++ b/metadata_service/proxy/neo4j_proxy.py
@@ -69,10 +69,6 @@ class Neo4jProxy(BaseProxy):
 
         wmk_results, table_writer, timestamp_value, owners, tags, source = self._exec_table_query(table_uri)
 
-        # neo4j returns is_view (boolean data type) as str, so we need to convert it to bool
-        is_view = self._safe_get(last_neo4j_record, 'tbl', 'is_view')
-        is_view = is_view.lower() in ['t', 'true'] if is_view else False
-
         table = Table(database=last_neo4j_record['db']['name'],
                       cluster=last_neo4j_record['clstr']['name'],
                       schema=last_neo4j_record['schema']['name'],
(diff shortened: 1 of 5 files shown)


metadata_service/proxy/neo4j_proxy.py
```diff
@@ -81,7 +81,8 @@ def get_table(self, *, table_uri: str) -> Table:
                       watermarks=wmk_results,
                       table_writer=table_writer,
                       last_updated_timestamp=timestamp_value,
-                      source=source)
+                      source=source,
+                      is_view=last_neo4j_record['tbl']['is_view'])
```
jinhyukchang: Would this break if the underlying node attribute does not have `is_view`?

You may use https://github.com/lyft/amundsenmetadatalibrary/blob/master/metadata_service/proxy/neo4j_proxy.py#L227

youngyjd: Good catch! 👍

youngyjd:
@feng-tao @jinhyukchang PTAL

youngyjd:
🔒

youngyjd:
🔓

metadata_service/proxy/neo4j_proxy.py
```diff
@@ -69,6 +69,10 @@ def get_table(self, *, table_uri: str) -> Table:
 
         wmk_results, table_writer, timestamp_value, owners, tags, source = self._exec_table_query(table_uri)
 
+        # neo4j returns is_view (boolean data type) as str, so we need to convert it to bool
+        is_view = self._safe_get(last_neo4j_record, 'tbl', 'is_view')
+        is_view = is_view.lower() in ['t', 'true'] if is_view else False
```
feng-tao: is there any better way?

youngyjd:
@feng-tao @jinhyukchang Please take a look

PR #31, opened 2019-04-29
Popular table query improvements

verdan:
This PR updates the `atlasclient` version, which is a big change, along with the queries improvements for popular tables. 
There used to be multiple calls in older version of atlasclient, which is fixed now. Also bumped the version of amundsen metadata.


verdan:
@feng-tao @jinhyukchang ready for review

verdan:
Sorry, I reverted the version back to `1.0.5` as there are some changes needed to be done on atlas_proxy based on new version of atlas. Will update the version after implementing those changes. 
@feng-tao

feng-tao:
Sounds good @verdan

Issue #32, opened 2019-04-30
Make amundsen compatible with upcoming release of Atlas i.e., v2.0


verdan:
Atlas is working on new release, which might break our implementation of Amundsen Atlas proxy. Make sure atlas's 1.1x+ versions are supported by amundsen. Keep backward compatibility in mind while implementing this.

verdan:
One of the breaking change is that the new version of Atlas does not include `relationshipAttributes` under `attribute` key anymore.

jinhyukchang:
@verdan Just reviewed the PR, merged and released it. Let me know if you need anything. Thanks!

verdan:
thanks @jinhyukchang. this can be closed for now :)

PR #33, opened 2019-04-30
Support for upcoming version of Apache Atlas

verdan:
This PR fixes following issue: https://github.com/lyft/amundsenmetadatalibrary/issues/32

Also bumps the version of metadata, since we recently made the queries efficient and made the atlas attributes more configurable based on user's deployments of Apache Atlas.

### DIFF
--- a/metadata_service/config.py
+++ b/metadata_service/config.py
@@ -27,6 +27,17 @@ class Config:
 
     IS_STATSD_ON = False
 
+    # Used to differentiate tables with other entities in Atlas. For more details:
+    # https://github.com/lyft/amundsenmetadatalibrary/blob/master/docs/proxy/atlas_proxy.md
+    ATLAS_TABLE_ENTITY = 'Table'
+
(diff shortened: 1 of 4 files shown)


verdan:
@feng-tao @jinhyukchang ready for review

metadata_service/config.py
```diff
@@ -27,6 +27,17 @@ class Config:
 
     IS_STATSD_ON = False
 
+    # Used to differentiate tables with other entities in Atlas. For more details:
+    # https://github.com/lyft/amundsenmetadatalibrary/blob/master/docs/proxy/atlas_proxy.md
+    ATLAS_TABLE_ENTITY = 'Table'
```
jinhyukchang: Are these something that needs to be configurable? Or is this the one expected to change in upcoming version?

verdan: These are the settings that should be configurable, as it can vary based on the atlas deployments and configurations within an organization.

Issue #22, opened 2019-04-09
Use entity IDs where possible instead of using the table URI


verdan:
For most of the part, amundsen metadata relies on the `table_uri`, which is being passed from frontend, that is making it complex and extra processing in case of systems which supports IDs. 
I am proposing to use the IDs of entities like Table, Column, and User etc to be used for the different kinds of operations. This information can be passed from metadata (when available, None otherwise), and simply be returned from frontend for each operation. 
Example Use Case: 
- Get Column Description:
At the moment, the frontend is sending the table_uri and the column name. I need to parse the table_uri, find table details, find the column based on the column name, and then get the column detail entity, and return the description. If there would be ID available for the column, then we can simply use that ID to get the record, and return the description.

Same is the case for Put, and Table detail itself. If we'd have the ID, we won't need to parse the table_uri. 

Just out of curiosity, how does Neo4j handles the IDs, does it support IDs ?

bolkedebruin:
Cc @feng-tao

jinhyukchang:
@verdan Neo4j supports ID but it's for internal purpose, not recommend for application to use.
Neo4j key (table URI for table node key) is defined in Databuilder and mostly shows its hierarchy. Thus,  column key can be easily derived from table key (table uri).

@verdan Could Atlas column can be derived from table key? Also, would you let me know where would be the proper place for column IDs to be available? (Don't think table detail API should ask for this and wonder where it can retrieve)

verdan:
@jinhyukchang no, there is no way I can derive column ID from Table ID/ Table Key. The path to get column information from table_uri is already stated in the description of this issue. 

P.S: Atlas uses GUIDs as the entity IDs.

bolkedebruin:
@verdan what about generating the qualifiedName from the URI? Ie qualifiedName for a column would be database.table.column@cluster which is indexed in atlas. Would that work?

verdan:
@bolkedebruin that is a possibility, but in that case we need to make it a standard for all the users who want to connect atlas with amundsen to name their columns in this way. 
And, what about the existing data in atlas?

jinhyukchang:
@verdan I can also see the benefit of using column entity ID as it can remove the knowledge of how it's being constructed from `metadata` service.

Table detail API to include column entity ID -> Frontend uses it when it calls column related API.

cc: @feng-tao

Whazor:
[...]

PR #42, opened 2019-05-08
Use name instead of qualifiedName and add cluster

Whazor:
### Summary of Changes

Currently the qualifiedName is the `normal name`@`cluster name` which makes all the names like `column@cluster1`.

Also, we switched to searching the table with `search_dsl`. Which allows us to search on all the properties (cluster, database, table) instead of only table name.

### Tests

I modified the atlas testing to adopt the new `search_dsl` change, also I added some patching and used `name` instead of `qualifiedName`.

### Documentation

No need to change documentation, will work by default with Atlas docker compose. No required settings change.

### CheckList
- [x] PR passes `make test`

### DIFF
--- a/metadata_service/config.py
+++ b/metadata_service/config.py
@@ -36,7 +36,7 @@ class Config:
 
     # Display name of Atlas Entities that we use for amundsen project.
     # Atlas uses qualifiedName as indexed attribute. but also supports 'name' attribute.
-    ATLAS_NAME_ATTRIBUTE = 'qualifiedName'
+    ATLAS_NAME_ATTRIBUTE = 'name'
 
 
 class LocalConfig(Config):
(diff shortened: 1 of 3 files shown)


bolkedebruin:
Qualified name is indexed and name afaik is not. Why not remove everything after the last '@' or remove it by using a configurable cluster name?

bolkedebruin:
Or just use name as a displayName but not for reference?

feng-tao:
cc @verdan

verdan:
As already mentioned by @bolkedebruin query on name is not indexed/efficient, and also this is making things complex by introducing the dsl search here to get only one entity, when it can easily be fetched using unique and indexed attribute. 

Once we have the unique attribute/key for each entity https://github.com/lyft/amundsenmetadatalibrary/issues/22 we can use the name attribute as the displayName and will use the `key` for making queries. In case of neo4j it would be table_uri, and in case of atlas the key would be simply a `guid` of atlas entity. This makes things much simpler and efficient.

I think for now we can close this issue, and make the qualifiedName in atlas to set without cluster name.

Whazor:
I will adopt the pull request tomorrow to use qualifiedName without search and map the same URL structure to the same structure that Neo4J uses.

PR #43, opened 2019-05-09
In Atlas Proxy, use correct hive URL and display name

Whazor:
### Summary of Changes

* Construct qualifiedName from URL

* Use Atlas qualifiedName structure: `database.table@cluster`

* In frontend show table name instead of `qualifiedName`

### Tests

* Correct qualifiedName in test tables

* Use name instead of qualifiedName for Table and Column types

* Add mocking decorator that patches instead of overwriting Atlas Client

### Documentation

Works by default with Atlas, user does not need to change anything.

### CheckList
- [x] PR passes `make test`

### DIFF
--- a/metadata_service/config.py
+++ b/metadata_service/config.py
@@ -36,7 +36,7 @@ class Config:
 
     # Display name of Atlas Entities that we use for amundsen project.
     # Atlas uses qualifiedName as indexed attribute. but also supports 'name' attribute.
-    ATLAS_NAME_ATTRIBUTE = 'qualifiedName'
+    ATLAS_NAME_ATTRIBUTE = 'name'
 
 
 class LocalConfig(Config):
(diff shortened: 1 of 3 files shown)


bolkedebruin:
Im a bit worried about this, as you are constructing the qualifiedName on the fly. While this goes for Tables really well how do you see this for S3, People etc?

Whazor:
In #22 it is discussed to use IDs/GUIDs instead of URL's, but this is more long term. For other entities we would hopefully can use qualifiedName without having to rewrite. But how Amundsen is currently implemented, we will always have a separate Python method for each entity and always an URL that will need to be matched to an entity in Atlas. I think per entity we would have to check at the moment.

bolkedebruin:
Im not so sure how "long" term that is. I think the change is not that invasive and showing the qualifiedName for the time being is not that bad either.

feng-tao:
cc @verdan

verdan:
I would suggest to keep the qualifiedName like the way it is for now in the master branch, and implement the GUIDs/IDs for the entities as a next step of improvement. That would solve quite a lot of parsing and issues with the consistency with the search service.

PR #44, opened 2019-05-17
Use PopularTable model when return user resource

feng-tao:
### Summary of Changes

When building Amundsen people, we found that it is not necessary to use Table model when return table object to FE. The table model contains all the metadata related table detail page which the payload size is huge. Switch to use popularTable model.

### Tests

### Documentation

### CheckList
- [x] PR passes `make test`

### DIFF
--- a/metadata_service/api/user.py
+++ b/metadata_service/api/user.py
@@ -3,7 +3,7 @@ from typing import Iterable, Mapping, Union
 
 from flask_restful import Resource, fields, marshal
 
-from metadata_service.api.table import table_detail_fields
+from metadata_service.api.popular_tables import popular_table_fields
 from metadata_service.exception import NotFoundException
 from metadata_service.proxy import get_proxy_client
 from metadata_service.util import UserResourceRel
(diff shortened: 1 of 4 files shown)


feng-tao:
cc @danwom @jinhyukchang

metadata_service/api/user.py
```diff
@@ -3,7 +3,7 @@
 
 from flask_restful import Resource, fields, marshal
 
-from metadata_service.api.table import table_detail_fields
+from metadata_service.api.popular_tables import popular_table_fields
```
danwom: Can you rename `popular_table_fields` to something generic like `table_summary_fields` and move it to a common file for tables in general?

danwom: Also I think `popular_table_fields` doesn't include `last_updated_epoch`. I think the data model used by search service is more complete.

feng-tao: this has been used by atlas proxy as well which we could do but will be in a different pr.

PR #45, opened 2019-05-17
Fix setup.py GitHub url

jornh:
Also tweaked description to be similar to search service

Currently if you go to pypi and click the homepage link you’ll get a 404


feng-tao:
thanks @jornh

PR #56, opened 2019-06-19
Pass table GUID in Atlas proxy and table URI in Neo4j proxy to 'key' respectively

L-KID:
### Summary of Changes

* Use  <code>key</code> instead of <code>table_uri</code> or <code>table_id</code> for all proxies
* Move <code>marshall_table_partial</code> to metadata
* Add instance attributes <code>key</code>, <code>entity_type</code> and <code>last_updated_epoch</code> to <code>PopularTable</code>
* Pass table GUID to <code>key</code> in Atlas, and construct table URI as <code>key</code> in Neo4j

### Tests

* Use key instead of table_uri in tests
* Correct PopularTable in tests

### Documentation

### CheckList
- [ ] PR adds unit tests, updates existing unit tests, __OR__ documents why no test additions or modifications are needed.
- [ ] PR passes `make test`
- [ ] I have squashed multiple commits if they address the same issue. In addition, my commits follow the guidelines from "[How to write a good git commit message](http://chris.beams.io/posts/git-commit/)"

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,13 +10,21 @@ script:
 after_success:
 - codecov
 deploy:
-  provider: pypi
+- provider: pypi
   user: amundsen-dev
   password:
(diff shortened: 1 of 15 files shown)


feng-tao:
I am not sure if I get the benefit on calling table_uri to key(cc @verdan). The table_uri is the protocol between FE and metadata. Unless there is a compelling reason, I would prefer to keep the existing name.

My understanding is that what . you need to do a transformation between table_uri to table_id in atlas proxy?

feng-tao:
FYI, there is no UUID concept in neo4j proxy.

verdan:
@feng-tao @L-KID 
To be able to support both proxies (atlas and Neo4j), we need to name the unique identifier something more meaningful. In case of atlas the unique identifier is `GUID` and for Neo4j it is `table_uri`. That's why Xin is trying to implement a solution that would serve both. 

So the flow on metadata side would be:
Neo4j:
- table_key: table_uri (will be used only for interactions b/w frontend and metadata)
- parse table key in metadata service, and send schema, db etc. from metadata proxy to frontend as payload. 

Atlas:
- table_key: Atlas GUID (will be used only for interactions b/w frontend and metadata)
- Pass the schema, db etc from metadata proxy to frontend as payload. 

P.S: I would call it `table_key`, `column_key` (or `table_id`, `column_id`) etc instead of just `key` to make things clear.

PR #59, opened 2019-07-01
Improve performance of getting bookmarked tables, refactored getting freq read table query

jinhyukchang:
### Summary of Changes

Previously, Metadata service was performing queries against each columns unnecessarily while getting bookmarked table. This change is to remove those query against columns.

Additionally, refactored query that retrieves frequently used table to fetch the one with `published_tag` and to get latest read data up to 50 tables.

### Tests

Unit tested. Integration tested locally.

### Documentation

N/A

### CheckList
- [x] PR passes `make test`

### DIFF
--- a/metadata_service/api/user.py
+++ b/metadata_service/api/user.py
@@ -121,7 +121,7 @@ class UserFollowAPI(Resource):
             LOGGER.exception('UserFollowAPI DELETE Failed')
             return {'message': 'The user {} for table_uri {} '
                                'is not deleted successfully'.format(user_id,
-                                                                  table_uri)}, \
+                                                                    table_uri)}, \
                 HTTPStatus.INTERNAL_SERVER_ERROR
 
 
--- a/metadata_service/api/user.py
+++ b/metadata_service/api/user.py
@@ -75,7 +75,7 @@ class UserFollowAPI(Resource):
             return {'message': 'user_id {} does not exist'.format(user_id)}, HTTPStatus.NOT_FOUND
 
         except Exception:
-            LOGGER.exception('UserFollowAPI Failed')
+            LOGGER.exception('UserFollowAPI GET Failed')
             return {'message': 'Internal server error!'}, HTTPStatus.INTERNAL_SERVER_ERROR
 
     def put(self, user_id: str, resource_type: str, table_uri: str) -> Iterable[Union[Mapping, int, None]]:
@@ -95,6 +95,7 @@ class UserFollowAPI(Resource):
                                'is added successfully'.format(user_id,
                                                               table_uri)}, HTTPStatus.OK
         except Exception as e:
+            LOGGER.exception('UserFollowAPI PUT Failed')
             return {'message': 'The user {} for table_uri {} '
                                'is not added successfully'.format(user_id,
                                                                   table_uri)}, \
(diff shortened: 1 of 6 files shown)


metadata_service/proxy/neo4j_proxy.py
```diff
@@ -752,33 +751,28 @@ def get_table_by_user_relation(self, *, user_email: str, relation_type: UserReso
...
         relation, _ = self._get_relation_by_type(relation_type)
-        # relationship can't be parameterized
-        query_key = 'key: "{user_id}"'.format(user_id=user_email)
 
         query = textwrap.dedent("""
-        MATCH (user:User {{{key}}})-[:{relation}]->(tbl:Table)
-        RETURN COLLECT(DISTINCT tbl) as table_records
-        """).format(key=query_key,
-                    relation=relation)
+MATCH (user:User {{key: $query_key}})-[:{relation}]->(tbl:Table)-[:TABLE_OF]->
```
feng-tao: the change looks good, but is there a way to extend this query to support the frequent use use case?
Here is my initial query I have in mind
```
MATCH (user:User {key: 'tfeng@lyft.com'})-[r:READ]->(tbl:Table)
WITH tbl.key as table_key, r.read_count as total_reads, r.published_tag as published_tag
where published_tag IS NOT NULL
RETURN  table_key as table_records, total_reads, published_tag  order by published_tag desc, total_reads desc limit 50;
```

jinhyukchang: Mmm. We might can squeeze it to make it shareable, but I am afraid it would make it complicated. How about separating it out?

feng-tao: Yeah. I think separate it out will be fine. Do you think you could combine that into this pr so that we could ship the change together? Thanks :)

feng-tao: sync offline. Let's push the change and add a todo on the query .

jinhyukchang: @feng-tao Just added `get_frequently_used_tables` method for frequently used table. :)

metadata_service/api/user.py
```diff
@@ -192,12 +204,12 @@ def get(self, user_id: str) -> Iterable[Union[Mapping, int, None]]:
         :return:
         """
         try:
-            resources = self.client.get_table_by_user_relation(user_email=user_id,
-                                                               relation_type=UserResourceRel.read)
+            resources = self.client.get_frequently_used_tables(user_email=user_id)
```
feng-tao: hey @jinhyukchang , you need to add this call to base_proxy and atlas_proxy as well.

jinhyukchang: Got it and added.

jinhyukchang:
> @jinhyukchang thanks for adding this. You need to add the new API call to base_proxy/atlas_proxy as well.

Done

feng-tao:
looks good! Have you tried to push a pre-release pypi and tested in staging?

jinhyukchang:
> looks good! Have you tried pushing a pre-release pypi and tested in staging?

I've tested out locally by calling our staging Neo4j. Should be good enough but will test out in staging again once released.

PR #63, opened 2019-07-16
[WIP]: POC for using key and parsing of keys on metadata proxy level

verdan:
### Summary of Changes

_Include a summary of changes then remove this line_

### Tests

### Documentation

### CheckList
- [ ] PR adds unit tests, updates existing unit tests, __OR__ documents why no test additions or modifications are needed.
- [ ] PR passes `make test`
- [ ] I have squashed multiple commits if they address the same issue. In addition, my commits follow the guidelines from "[How to write a good git commit message](http://chris.beams.io/posts/git-commit/)"

### DIFF
--- a/metadata_service/api/popular_tables.py
+++ b/metadata_service/api/popular_tables.py
@@ -7,6 +7,7 @@ from flask_restful import Resource, fields, marshal
 from metadata_service.proxy import get_proxy_client
 
 popular_table_fields = {
+    'key': fields.String,
     'database': fields.String,
     'cluster': fields.String,
     'schema': fields.String,
--- a/metadata_service/config.py
+++ b/metadata_service/config.py
@@ -37,7 +37,7 @@ class Config:
     # FixMe: Once GUIDs are in place, then change it to 'name' (display name)
     # Display name of Atlas Entities that we use for amundsen project.
     # Atlas uses qualifiedName as indexed attribute. but also supports 'name' attribute.
-    ATLAS_NAME_ATTRIBUTE = 'qualifiedName'
+    ATLAS_NAME_ATTRIBUTE = 'name'
 
 
 class LocalConfig(Config):
--- a/metadata_service/entity/popular_table.py
+++ b/metadata_service/entity/popular_table.py
@@ -6,6 +6,7 @@ from marshmallow_annotations.ext.attrs import AttrsSchema
 
 @attr.s(auto_attribs=True, kw_only=True)
 class PopularTable:
+    key: str = attr.ib()
     database: str = attr.ib()
     cluster: str = attr.ib()
     schema: str = attr.ib()
--- a/metadata_service/entity/popular_table.py
+++ b/metadata_service/entity/popular_table.py
@@ -4,11 +4,13 @@ from typing import Optional
 class PopularTable:
 
     def __init__(self, *,
+                 key: str,
                  database: str,
                  cluster: str,
                  schema: str,
                  name: str,
                  description: Optional[str] = None) -> None:
+        self.key = key
         self.database = database
         self.cluster = cluster
         self.schema = schema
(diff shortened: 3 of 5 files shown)


verdan:
@jinhyukchang @feng-tao this is what we are suggesting to do with using the `key` for each entity/model and use that on the frontend (https://github.com/lyft/amundsenmetadatalibrary/issues/22). Keep all the logic in metadata. The frontend would be like this: https://github.com/lyft/amundsenfrontendlibrary/pull/227

verdan:
@ttannis @danwom FYI ^

danwom:
I am 110% in favor of using a single `key` or `id` field for all resource entities. 

We currently use the `/` character as a delimiter in the `key`, which is also the delimiter for URL route params. This makes it difficult to parse the `key` as a single route.

```<Route path="/table_detail/:cluster/:db/:schema/:table" component={TableDetail} />```
vs 
```<Route path="/table_detail/:table_key" component={TableDetail} />```
 https://github.com/lyft/amundsenfrontendlibrary/pull/227/files#diff-30061a6dfc2c2259725200dad29d98c0R41. 


URI encoding/decoding will create ugly URLs but should work. 
`/table_detail/cluster/db/schema/table` -> `/table_detail/cluster%2Fdb%2Fschema%2Ftable`

Another solution would be to use a different delimiting character for keys e.g. `cluster:db_name:schema:table` or `cluster#db_name#schema#table`

verdan:
> I am 110% in favor of using a single `key` or `id` field for all resource entities.
> 
> We currently use the `/` character as a delimiter in the `key`, which is also the delimiter for URL route params. This makes it difficult to parse the `key` as a single route.
> 
> `<Route path="/table_detail/:cluster/:db/:schema/:table" component={TableDetail} />`
> vs
> `<Route path="/table_detail/:table_key" component={TableDetail} />`
> https://github.com/lyft/amundsenfrontendlibrary/pull/227/files#diff-30061a6dfc2c2259725200dad29d98c0R41.
> 
> URI encoding/decoding will create ugly URLs but should work.
> `/table_detail/cluster/db/schema/table` -> `/table_detail/cluster%2Fdb%2Fschema%2Ftable`
> 
> Another solution would be to use a different delimiting character for keys e.g. `cluster:db_name:schema:table` or `cluster#db_name#schema#table`

For Atlas, I am planning to use GUID as key, that will make it `table_detail/<GUID>`. It will also help us working with different versions. I don't really mind having ugly urls (meaningless GUID in URL). It will make things pretty straight forward and easy with the GUIDs in URL. @bolkedebruin do you have any other thoughts about this? 

I also like the idea of using colons or hashes. @feng-tao your thoughts? Once decided, I can also make the change for neo4j proxy.

bolkedebruin:
@verdan I don't have a particular opinion on this. As readable urls might be nice for deeplinking it could be an idea to support both, but not a strong requirement IMHO.

ttannis:
@verdan we will need to maintain the option to have the human readable urls in amundsenfrontendlibrary even if that's not what you will be using, so whatever changes will be made here for the neo4j proxy and planned for lyft/amundsenfrontendlibrary#227, we'll have to make sure that both options are possible

feng-tao:
late for the game, but agree that we need to maintain the user readable URL as internally we have so many different users bookmark different URLs . Any URL backward incompatible won't work in our case.

Another question would be how to handle to existing data migration if we choose to move to a new format. 

cc @jinhyukchang

verdan:
Oops, sorry I totally missed the bookmarks feature. (and to maintain the backward compatibility because of that). 
I will think about it a little more, and open the PR with a better solution. Thanks for the feedback.

feng-tao:
@verdan no worry. Thanks always for the contribution. One question though : is it ING not able to use the new bookmark feature because the FE hasn't been updated its pypi version for a while?

verdan:
@feng-tao we are doing a lot of research in terms of data quality, lineage etc., and also fixing/adding new features in related open source projects as well, which is keeping us a bit behind in terms of Amundsen. 
About bookmarks, we haven't added the support of user profile and bookmarks in atlas proxy yet.

PR #64, opened 2019-07-16
Use attrs + marshmallow-annotations for entities

friendtocephalopods:
### Summary of Changes

Use attrs to generate python classes and marshmallow-annotations to generate schemas.

This makes it far, far less painful to add new entities.

### Tests

None modified. Refactor.

### Documentation

None modified. Refactor.

### CheckList
- [ x ] PR title addresses the issue accurately and concisely. Example: "Updates the version of Flask to v1.0.2"
- [ x ] PR includes a summary of changes. 
- [ x ] PR adds unit tests, updates existing unit tests, __OR__ documents why no test additions or modifications are needed.
- [ x ] In case of new functionality, my PR adds documentation that describes how to use it.
- [ x ] PR passes `make test`
- [ x ] I have squashed multiple commits if they address the same issue. In addition, my commits follow the guidelines from "[How to write a good git commit message](http://chris.beams.io/posts/git-commit/)"

### DIFF
--- a/metadata_service/entity/table_detail.py
+++ b/metadata_service/entity/table_detail.py
@@ -1,175 +1,176 @@
-from typing import Iterable, Optional
+from typing import List, Optional
+
+import attr
+from marshmallow_annotations.ext.attrs import AttrsSchema
 
 
+@attr.s(auto_attribs=True, kw_only=True)
 class User:
-    def __init__(self, *,
-                 email: str,
-                 first_name: str =None,
-                 last_name: str =None) -> None:
-        self.email = email
-        self.first_name = first_name
-        self.last_name = last_name
-
-    def __repr__(self) -> str:
-        return 'User(email={!r}, first_name={!r}, last_name={!r})'.format(self.email, self.first_name, self.last_name)
+    email: str = attr.ib()
+    first_name: Optional[str] = None
+    last_name: Optional[str] = None
 
 
+class UserSchema(AttrsSchema):
+    class Meta:
+        target = User
+        register_as_scheme = True
+
+
+@attr.s(auto_attribs=True, kw_only=True)
 class Reader:
-    def __init__(self, *,
-                 user: User,
-                 read_count: int) -> None:
-        self.user = user
-        self.read_count = read_count
-
-    def __repr__(self) -> str:
-        return 'Reader(user={!r}, read_count={!r})'.format(self.user, self.read_count)
+    user: User = attr.ib()
+    read_count: int = attr.ib()
 
 
+class ReaderSchema(AttrsSchema):
+    class Meta:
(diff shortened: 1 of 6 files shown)


feng-tao:
thanks @friendtocephalopods for your contributions!. Will take a look. cc @jinhyukchang as well

jinhyukchang:
@friendtocephalopods I am not familiar with `marshmallow-annotations` and could you explain the benefit of it?

cc: @bolkedebruin @verdan

friendtocephalopods:
Sure! Marshmallow-annotations is able to generate schemas using type hints, so you don't need to define schemas by hand.

In this case, the schemas would be useful for the common amundsen-models/amundsen-entities package that's forthcoming, as well as a PUT/POST tables endpoint for metadatalibrary I'd be happy to create a separate PR for

bolkedebruin:
O thats interesting! It could be useful for airflow as well if so we should put this in a separate package maybe (ie the schema definitions). The meta stuff attracts me less ;-)

verdan:
This is actually quite nice, however, I'd rather move the models to a common place first, and then implement annotator there. It seems like you'll need to make the change in 3 places in order to make it consistent at the moment, which is not ideal. 
How far are we in planning to move models/entities in a common repo @feng-tao @jinhyukchang ?

feng-tao:
@verdan, @friendtocephalopods  . Make sense. @jinhyukchang is working with our OSS team to open source that common repo. Currently it is review cycle, will get back to this once it is finished. Stay tuned!

jinhyukchang:
> This is actually quite nice, however, I'd rather move the models to a common place first, and then implement annotator there. It seems like you'll need to make the change in 3 places in order to make it consistent at the moment, which is not ideal.
> How far are we in planning to move models/entities in a common repo @feng-tao @jinhyukchang ?

Good idea @verdan . `amundsen-common` repo just initiated procedure of open sourcing it and I expect it will be done in few weeks but at the same time we don't have control as depends on few external teams within Lyft.

I suggest to adopt this PR and have a separate step to place it in common repo. What do you think?

verdan:
sounds good to me @jinhyukchang 👍 
cc: @feng-tao

metadata_service/entity/table_detail.py
```diff
@@ -1,175 +1,176 @@
...
+        register_as_scheme = True
+
+
+@attr.s(auto_attribs=True, kw_only=True)
+class PiiDetail:
+    pii_semantic_type: Optional[str] = None
+    pii_forget_type: Optional[str] = None
+
 
+class PiiDetailSchema(AttrsSchema):
```
feng-tao: do we use this new model anywhere?

PR #75, opened 2019-08-27
Added cache for atlas proxy popular table API.

morpheyesh:
### Summary of Changes

* This PR includes caching of popular tables API for atlas proxy.
* Implementation is similar to neo4j's and uses beaker.
* Added _get_metadata_collection function which isolates the two atlas calls for fetching metadata id and metadata collection.

### Tests

* Changed test cases in `test_atlas_proxy.py`

### Documentation

Nothing was added or removed.

### CheckList
- [ *] PR title addresses the issue accurately and concisely. Example: "Updates the version of Flask to v1.0.2"
- [*] PR includes a summary of changes. 
- [*] PR adds unit tests, updates existing unit tests, __OR__ documents why no test additions or modifications are needed.
- [*] In case of new functionality, my PR adds documentation that describes how to use it.
- [*] PR passes `make test`
- [*] I have squashed multiple commits if they address the same issue. In addition, my commits follow the guidelines from "[How to write a good git commit message](http://chris.beams.io/posts/git-commit/)"

### DIFF
--- a/metadata_service/proxy/atlas_proxy.py
+++ b/metadata_service/proxy/atlas_proxy.py
@@ -19,12 +19,14 @@ from metadata_service.exception import NotFoundException
 from metadata_service.proxy import BaseProxy
 from metadata_service.util import UserResourceRel
 
-_CACHE = CacheManager(**parse_cache_config_options({'cache.type': 'memory'}))
+LOGGER = logging.getLogger(__name__)
 
 # Expire cache every 11 hours + jitter
-_GET_POPULAR_TABLE_CACHE_EXPIRY_SEC = 11 * 60 * 60 + randint(0, 3600)
+_ATLAS_PROXY_CACHE_EXPIRY_SEC = 11 * 60 * 60 + randint(0, 3600)
 
-LOGGER = logging.getLogger(__name__)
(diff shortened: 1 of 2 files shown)


metadata_service/proxy/atlas_proxy.py
```diff
@@ -325,24 +333,30 @@ def get_column_description(self, *,
...
+        :param num_entries: Number of popular tables to fetch
+        :return: A List of popular tables instances
+        """
+        popular_tables = list()
+        query_metadata_ids = {'query': f'FROM Table SELECT metadata.__guid '
+                                       f'ORDERBY popularityScore desc '
+                                       f'LIMIT {num_entries}'}
+
+        metadata_collection = self._get_metadata_collection(query_metadata_ids)
+
```
verdan: Thanks @morpheyesh, this is really nice. 
I'd suggest to move the main getting entities logic to the cache function.
Like instead of _get_metadata_collection, your function should be _get_metadata_entities.
because essentially there would always be one item in the collection.

because this function `_collection.entities_with_relationships(attributes=["parentEntity"])` also need to make some queries to Atlas, which we don’t really want everytime.

feng-tao:
thanks @verdan for the review!

PR #79, opened 2019-09-19
Improve table & columns description formatting (#98)

Mikhail-Ivanov:
### Summary of Changes

Introduced support of Markdown descriptions for tables & columns.
This module changes are intended to pass description text as a request body (instead of using URI parameter).

[Related changes at amundsenfrontendlibrary.](https://github.com/lyft/amundsenfrontendlibrary/pull/298)

### Tests

Unit tests for table and column description upsertion have been modified

### Documentation

Markdown description format demo:
https://rexxars.github.io/react-markdown/

### DIFF
--- a/metadata_service/api/column.py
+++ b/metadata_service/api/column.py
@@ -21,16 +21,18 @@ class ColumnDescriptionAPI(Resource):
 
     def put(self,
             table_uri: str,
-            column_name: str,
-            description_val: str) -> Iterable[Union[dict, tuple, int, None]]:
+            column_name: str) -> Iterable[Union[dict, tuple, int, None]]:
         """
-        Updates column description
+        Updates column description (passed as a request body)
+        :param table_uri:
+        :param column_name:
+        :return:
         """
         try:
+            description = self.parser.parse_args()['description']
             self.client.put_column_description(table_uri=table_uri,
                                                column_name=column_name,
-                                               description=description_val)
-
+                                               description=description)
             return None, HTTPStatus.OK
 
         except NotFoundException:
--- a/metadata_service/api/column.py
+++ b/metadata_service/api/column.py
@@ -1,7 +1,9 @@
+import json
 from http import HTTPStatus
 from typing import Iterable, Union
 
-from flask_restful import Resource, reqparse
+from flask import request
+from flask_restful import Resource
 
 from metadata_service.exception import NotFoundException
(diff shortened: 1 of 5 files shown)


feng-tao:
Seems pretty safe. In the mean time, could you also fix the CLA check? Normally it is due to you didn't configure your github handle(username, email) in your local box.
cc @jinhyukchang

metadata_service/api/column.py
```diff
@@ -21,16 +21,18 @@ def __init__(self) -> None:
...
+            column_name: str) -> Iterable[Union[dict, tuple, int, None]]:
         """
-        Updates column description
+        Updates column description (passed as a request body)
+        :param table_uri:
+        :param column_name:
+        :return:
         """
         try:
+            description = self.parser.parse_args()['description']
```
feng-tao: @verdan , I assume it won't affect the atlas backend?

verdan: It shouldn't. go ahead!!

feng-tao:
the backend code looks fine. So once we are agree with the FE part, we are good to go with this pr. And please increase the version as well.

Mikhail-Ivanov:
Please note that I've got rid of reqparse because of this (also faced issues while writing unit tests):
https://stackoverflow.com/questions/19384526/how-to-parse-the-post-argument-to-a-rest-service

metadata_service/api/column.py
```diff
@@ -13,24 +15,22 @@ class ColumnDescriptionAPI(Resource):
...
+            column_name: str) -> Iterable[Union[dict, tuple, int, None]]:
         """
-        Updates column description
+        Updates column description (passed as a request body)
+        :param table_uri:
+        :param column_name:
+        :return:
         """
         try:
+            description = json.loads(request.json).get('description')
```
ttannis: What do you all think about having a check for what happens if the description is `None`? Does `self.client.put_column_description` end up handling this, or will the description be deleted?

Mikhail-Ivanov: I've checked it locally (via Postman):
1) {"description": ""} - this request payload will truncate description
2) request payload with no "description" tag ends with {"msg": "Encountered exception: None"} and description data stay untouched

ttannis: Thanks for confirming the behavior for empty string and `None`. They behave as we expect, so no changes would be needed.

PR #80, opened 2019-09-23
optimize docker

jornh:
### Summary of Changes

Optimize docker - see https://github.com/lyft/amundsen/issues/109

### Tests

Manually tested locally with `docker build ...`

### Documentation

Should be on par with the Search service now

### CheckList
- [x] PR passes `make test`

### DIFF
--- a/public.Dockerfile
+++ b/public.Dockerfile
@@ -1,8 +1,11 @@
-FROM python:3
+FROM python:3-slim
 WORKDIR /app
-COPY . /app
-RUN pip3 install -r requirements.txt
-RUN python3 setup.py install
 RUN pip3 install gunicorn
+
+COPY requirements.txt /app/requirements.txt
+RUN pip3 install -r requirements.txt
+
+COPY . /app
+RUN python3 setup.py install
 
 CMD [ "python3", "metadata_service/metadata_wsgi.py" ]
--- a/setup.py
+++ b/setup.py
@@ -2,7 +2,7 @@ import os
 
 from setuptools import setup, find_packages
 
-__version__ = '1.0.19'
+__version__ = '1.1.0'
 
 requirements_path = os.path.join(os.path.dirname(os.path.realpath(__file__)), 'requirements.txt')
 with open(requirements_path) as requirements_file:
--- a/README.md
+++ b/README.md
@@ -2,11 +2,11 @@
 [![PyPI version](https://badge.fury.io/py/amundsen-metadata.svg)](https://badge.fury.io/py/amundsen-metadata)
 [![Build Status](https://api.travis-ci.com/lyft/amundsenmetadatalibrary.svg?branch=master)](https://travis-ci.com/lyft/amundsenmetadatalibrary)
 [![Coverage Status](https://img.shields.io/codecov/c/github/lyft/amundsenmetadatalibrary/master.svg)](https://codecov.io/github/lyft/amundsenmetadatalibrary?branch=master)
-[![License](http://img.shields.io/:license-Apache%202-blue.svg)](LICENSE)
-[![PRs Welcome](https://img.shields.io/badge/PRs-welcome-brightgreen.svg)](https://img.shields.io/badge/PRs-welcome-brightgreen.svg)
+[![License](https://img.shields.io/:license-Apache%202-blue.svg)](LICENSE)
+[![PRs Welcome](https://img.shields.io/badge/PRs-welcome-brightgreen.svg)](#developer-guide)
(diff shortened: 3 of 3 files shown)


public.Dockerfile
```diff
@@ -1,8 +1,11 @@
-FROM python:3
+FROM python:3-slim
```
ttannis: Asking for my own understanding, is the gist that `python:3-slim` makes this all faster? Any place where we can read more about it, I see the Airflow option uses another option `slim-stretch` and want to understand more about what this all does.  I did a Google search but I'm getting more hits about something referred to as `alpine`?

setup.py
```diff
@@ -2,7 +2,7 @@
 
 from setuptools import setup, find_packages
 
-__version__ = '1.0.19'
+__version__ = '1.1.0'
```
ttannis: Any reason that this kind of change becomes a minor release as opposed to a patch?

public.Dockerfile
jornh: As I’ve understood the base language `python`, `node` etc. Docker images are based on the Debian distribution. The more info you add to the version part after `:` the more specific you get `-slim` are trimmed down versions, but still based on Debian (examples are that they strip out base OS Python/Perl? etc versions and just add the one you give as “pinned down”). `3` is “current” Python 3, but not as specific as 3.6 etc. `-stretch`, `-buster` etc. reference Debian versions. 

`alpine` is different though, as it’s based on the Alpine distribution. It’s generally a bit smaller than Debian slim - but comes with it’s own Musl lib, so didn’t think it was worth the risk. Recently came across  https://pythonspeed.com/articles/base-image-python-docker-images/ which IMO generally sounds sensible and reflects my experiences with Docker quite well  (He’s also got other posts - mostly pythonic though ..)

I simply opted for `-slim` to reduce size, and thus (mostly) reduced waiting for download. For example with the search repo the uncompressed size is ~225 vs +900 MB. I didn’t add further version pinning down restrictions for now - but we can certainly discuss 

*(of interest for you and @danwom +other frontend folks I’m considering if it’s now time to bump Node to `-10` On what I’ll push after discussion has landed on the first two PRs ... but since you’re better qualified in that end this is the perfect time/place to ask 🙂 )*

setup.py
jornh: No particularly for these PRs it’s even superdebatable - because the changes only affect Docker users, but nothing new for PyPi package users. 

But I think Amundsen is generally erring on the “to many patches, to few minor changes”. Not sure if a policy was ever discussed (i.e. are we doing semantic versioning?) My understanding is patches are “no new features - just correction of bugs”. So if it does some new thing you bump minor. Happy to be enlightened though, not a super strong point for me!

So, for this particular change my thinking was: *”it adds a new feature to rely on for the compose file, so it’s a minor feature not just a fix/patch”*

feng-tao: I c, currently we don't doing semantic versioning as Lyft internally relies on import the library version to pick up the change. We could have a discussion later if semantic versioning is sth we aim for.

Also feel free to ignore my comment on the search pr. I think it make sense to bump the minor version per your comment.

public.Dockerfile
ttannis: I see. `-slim` seems good enough! 

Regrading bumping to `Node 10` -- let's continue the discussion in the slack channel if you want.

setup.py
jornh: 👍 that was a quick discussion! Agree it’s just numbers (and there are plenty to use + not that important if you always pin where you use stuff)

ttannis: > So, for this particular change my thinking was: _”it adds a new feature to rely on for the compose file, so it’s a minor feature not just a fix/patch”_

I can see that. Sounds good. Yes we're using semantic versioning.

ttannis: Hm, okay guess we're not using semantic versioning here. 😅 It's what we're aiming for on amundsenfrontendlibrary though, because if we don't make the distinction of bug vs new feature vs backwards incompatible change on the frontend -- people will end up getting really confused and unable to get the app to work. Community meeting/slack channel topic?

public.Dockerfile
jornh: Sure, that thing is certainly worth opening up to to the wider community!

setup.py
jornh: 😆 there! Now I started getting you Lyft’ers all confused about what you’re *actually doing* (honestly not my intention, I swear) If y’all up for it we can certainly open it up OTOH we should  watch out for the “semantic police” (though I’ve only seen sensible people around here so far) ... I’ll let you Lyft folks decide and take the lead on it.

PR #84, opened 2019-10-08
Adds unit tests

swatro:
### Summary of Changes
* Adds unit tests for the column description and the popular tables APIs.
* Excludes import statements from the test coverage report. This will change the code coverage on all files, but it will now be more reflective of code actually tested. 

### Tests
This mostly adds unit tests. 

### Documentation
None

### CheckList
- [x] PR passes `make test`

### DIFF
--- a/metadata_service/api/popular_tables.py
+++ b/metadata_service/api/popular_tables.py
@@ -27,6 +27,6 @@ class PopularTablesAPI(Resource):
         self.client = get_proxy_client()
 
     def get(self) -> Iterable[Union[Mapping, int, None]]:
-        limit = request.args.get('limit', 10)
+        limit = request.args.get('limit', 10, type=int)
         popular_tables = self.client.get_popular_tables(num_entries=limit)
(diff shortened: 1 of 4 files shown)


metadata_service/api/popular_tables.py
```diff
@@ -27,6 +27,6 @@ def __init__(self) -> None:
         self.client = get_proxy_client()
 
     def get(self) -> Iterable[Union[Mapping, int, None]]:
-        limit = request.args.get('limit', 10)
+        limit = request.args.get('limit', 10, type=int)
```
feng-tao: thanks. is this line change needed?

swatro: Without the type the `limit` variable would come in as a string instead of an integer. The function interface says it wants an integer. I believe the downstream function will work either way (the value is just being put into a query). 

I can remove it if you would prefer.

feng-tao: I think it is fine :) Thanks for the info.

PR #85, opened 2019-10-09
Unit tests for table

swatro:
### Summary of Changes
* Adds unit tests for the table APIs
* Removes '/table/<path:table_uri>/tag' API path as it does not work. 
* Removes unused code in the table API.
* Uses `BasicTestCase` as the best for other unit tests. 

### Tests
This mostly adds unit tests. 

### Documentation
None 
### CheckList
- [x] PR passes `make test`

### DIFF
--- a/metadata_service/api/table.py
+++ b/metadata_service/api/table.py
@@ -114,7 +114,7 @@ class TableOwnerAPI(Resource):
             return {'message': 'The owner {} for table_uri {} '
                                'is added successfully'.format(owner,
                                                               table_uri)}, HTTPStatus.OK
-        except Exception as e:
+        except Exception:
             return {'message': 'The owner {} for table_uri {} '
                                'is not added successfully'.format(owner,
                                                                   table_uri)}, HTTPStatus.INTERNAL_SERVER_ERROR
--- a/tests/unit/api/table/table_test_case.py
+++ b/tests/unit/api/table/table_test_case.py
@@ -1,0 +1,14 @@
+from mock import patch, Mock
+
+from tests.unit.test_basics import BasicTestCase
+
+
+class TableTestCase(BasicTestCase):
+    def setUp(self) -> None:
+        super().setUp()
+        self.mock_client = patch('metadata_service.api.table.get_proxy_client')
+        self.mock_proxy = self.mock_client.start().return_value = Mock()
+
+    def tearDown(self):
+        super().tearDown()
+        self.mock_client.stop()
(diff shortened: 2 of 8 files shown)


metadata_service/__init__.py
```diff
@@ -72,9 +72,7 @@ def create_app(*, config_module_class: str) -> Flask:
     api.add_resource(TableDetailAPI, '/table/<path:table_uri>')
     api.add_resource(TableDescriptionAPI,
                      '/table/<path:table_uri>/description')
-    api.add_resource(TableTagAPI,
-                     '/table/<path:table_uri>/tag',
```
swatro: I removed this path because it currently doesn't work. The `TableTagAPI` has two method `put` and `delete`. Both of the methods require two positional arguments (`table_uri` and `tag`). In order this path to work, we would need to remove the `tag` position argument (or make it optional?).

metadata_service/api/table.py
```diff
@@ -114,7 +113,7 @@ def put(self, table_uri: str, owner: str) -> Iterable[Union[Mapping, int, None]]
             return {'message': 'The owner {} for table_uri {} '
                                'is added successfully'.format(owner,
                                                               table_uri)}, HTTPStatus.OK
-        except Exception as e:
```
swatro: not using the `e` so why have it?

metadata_service/api/table.py
```diff
@@ -176,8 +175,6 @@ class TableTagAPI(Resource):
 
     def __init__(self) -> None:
         self.client = get_proxy_client()
-        self.parser = reqparse.RequestParser()
-        self.parser.add_argument('tag', type=str, location='json')
```
swatro: we never use this parser, so i deleted it. Happy to add it back if we need.

metadata_service/__init__.py
feng-tao: hey @swatro , i think the API is created for other services to push tags(Internally Lyft uses this API to push tag for ML feature table). Given the pr is for adding new unit test, I prefer to keep the API.

Sorry for documenting the path, the API is needed as metadata service not just serves FE but other services as well.

Let me know what you think.

swatro: You're right that this PR is really about unit tests. I'll put this code back. I do think there is a problem with this path not working. But I can put up a different PR for that.

swatro: @feng-tao I put up another [PR](https://github.com/lyft/amundsenmetadatalibrary/pull/86) to explain the problems with this API. I might be wrong, but I'm pretty sure only one path for this API is currently working

PR #91, opened 2019-12-04
Add badge tag type support

feng-tao:
Add badge support. Badge is non UI modifiable tag which allows other service to ingest.

Test with the following:
```tfeng@tfeng-mbp143 ~/src/amundsenmetadata-private (master) $ curl -X GET http://0.0.0.0:5000/table/"hive://gold.schema/table"/tag/?tag_type=badge
[{"tag_type": "badge", "tag_name": "test-tag1"}]
tfeng@tfeng-mbp143 ~/src/amundsenmetadata-private (master) $ curl -X PUT http://0.0.0.0:5000/table/"hive://gold.schema/table"/tag/test-tag2?tag_type=badge
{"message": "The tag test-tag2 for table_uri hive://gold.schema/table with type badge is not added successfully as badge is not part of the whitelist"}
tfeng@tfeng-mbp143 ~/src/amundsenmetadata-private (master) $ curl -X PUT http://0.0.0.0:5000/table/"hive://gold.schema/table"/tag/test-tag?tag_type=badge
{"message": "The tag test-tag for table_uri hive://gold.schema/table with type badge is added successfully"}
tfeng@tfeng-mbp143 ~/src/amundsenmetadata-private (master) $ curl -X GET http://0.0.0.0:5000/table/"hive://gold.schema/table"/tag/?tag_type=badge
[{"tag_type": "badge", "tag_name": "test-tag"}, {"tag_type": "badge", "tag_name": "test-tag1"}]
tfeng@tfeng-mbp143 ~/src/amundsenmetadata-private (master) $ curl -X DELETE http://0.0.0.0:5000/table/"hive://gold.schema/table"/tag/test-tag?tag_type=badge
{"message": "The tag test-tag for table_uri hive://gold.schema/table with type badge is deleted successfully"}```
it works as expected.

### DIFF
--- a/metadata_service/__init__.py
+++ b/metadata_service/__init__.py
@@ -73,7 +73,6 @@ def create_app(*, config_module_class: str) -> Flask:
     api.add_resource(TableDescriptionAPI,
                      '/table/<path:table_uri>/description')
     api.add_resource(TableTagAPI,
-                     '/table/<path:table_uri>/tag',
                      '/table/<path:table_uri>/tag/<tag>')
     api.add_resource(TableOwnerAPI,
                      '/table/<path:table_uri>/owner/<owner>')
--- a/metadata_service/__init__.py
+++ b/metadata_service/__init__.py
@@ -73,7 +73,6 @@ def create_app(*, config_module_class: str) -> Flask:
     api.add_resource(TableDescriptionAPI,
                      '/table/<path:table_uri>/description')
     api.add_resource(TableTagAPI,
-                     '/table/<path:table_uri>/tag/',
                      '/table/<path:table_uri>/tag/<tag>')
     api.add_resource(TableOwnerAPI,
                      '/table/<path:table_uri>/owner/<owner>')
--- a/metadata_service/__init__.py
+++ b/metadata_service/__init__.py
@@ -73,6 +73,7 @@ def create_app(*, config_module_class: str) -> Flask:
     api.add_resource(TableDescriptionAPI,
                      '/table/<path:table_uri>/description')
     api.add_resource(TableTagAPI,
+                     '/table/<path:table_uri>/tag/',
                      '/table/<path:table_uri>/tag/<tag>')
     api.add_resource(TableOwnerAPI,
                      '/table/<path:table_uri>/owner/<owner>')
--- a/metadata_service/api/table.py
+++ b/metadata_service/api/table.py
@@ -188,7 +188,7 @@ class TableTagAPI(Resource):
         tag_type = args.get('tag_type', 'default')
         try:
             tags = self.client.get_tag(table_uri=table_uri, tag_type=tag_type)
-            return marshal(tags, table_detail_fields), HTTPStatus.OK
+            return marshal(tags, tag_fields), HTTPStatus.OK
 
         except NotFoundException:
             return {'message': 'table_uri {} does not exist'.format(table_uri)}, HTTPStatus.NOT_FOUND
--- a/metadata_service/api/table.py
+++ b/metadata_service/api/table.py
@@ -195,14 +195,16 @@ class TableTagAPI(Resource):
             self.client.add_tag(table_uri=table_uri,
                                 tag=tag,
                                 tag_type=tag_type)
-            return {'message': 'The tag {} for table_uri {} '
+            return {'message': 'The tag {} for table_uri {} with type {} '
                                'is added successfully'.format(tag,
-                                                              table_uri)}, HTTPStatus.OK
+                                                              table_uri,
+                                                              tag_type)}, HTTPStatus.OK
         except NotFoundException:
             return \
-                {'message': 'The tag {} for table_uri {} '
+                {'message': 'The tag {} for table_uri {} with type {} '
                             'is not added successfully'.format(tag,
-                                                               table_uri)}, \
+                                                               table_uri,
+                                                               tag_type)}, \
                 HTTPStatus.NOT_FOUND
 
     def delete(self, table_uri: str, tag: str) -> Iterable[Union[Mapping, int, None]]:
@@ -219,13 +221,13 @@ class TableTagAPI(Resource):
             self.client.delete_tag(table_uri=table_uri,
                                    tag=tag,
                                    tag_type=tag_type)
-            return {'message': 'The tag {} for table_uri {} with type {}'
+            return {'message': 'The tag {} for table_uri {} with type {} '
                                'is deleted successfully'.format(tag,
                                                                 table_uri,
                                                                 tag_type)}, HTTPStatus.OK
         except NotFoundException:
             return \
-                {'message': 'The tag {} for table_uri {} with type {}'
+                {'message': 'The tag {} for table_uri {} with type {} '
                             'is not deleted successfully'.format(tag,
                                                                  table_uri,
                                                                  tag_type)}, \
--- a/metadata_service/api/table.py
+++ b/metadata_service/api/table.py
@@ -2,10 +2,11 @@ import json
 from http import HTTPStatus
 from typing import Iterable, Mapping, Union, Any
 
+from flask import current_app as app
 from flask import request
 from flask_restful import Resource, fields, reqparse, marshal
 
-from metadata_service.exception import NotFoundException, BadgeNotInWhitelistException
+from metadata_service.exception import NotFoundException
 from metadata_service.proxy import get_proxy_client
 
 
@@ -68,6 +69,7 @@ table_detail_fields = {
     'table_name': fields.String(attribute='name'),
     'table_description': fields.String(attribute='description'),  # Optional
     'tags': fields.List(fields.Nested(tag_fields)),  # Can be an empty list
+    'badges': fields.List(fields.Nested(tag_fields)),  # Can be an empty list
     # Can be an empty list
     'table_readers': fields.List(fields.Nested(table_reader_fields)),
     # Can be an empty list
@@ -180,22 +182,6 @@ class TableTagAPI(Resource):
         self.parser.add_argument('tag_type', type=str, required=False, default='default')
         super(TableTagAPI, self).__init__()
 
-    def get(self, table_uri: str) -> Iterable[Any]:
-        """
-        Returns all the tags based on the type in Neo4j endpoint
-        """
-        args = self.parser.parse_args()
-        tag_type = args.get('tag_type', 'default')
-        try:
-            tags = self.client.get_tag(table_uri=table_uri, tag_type=tag_type)
-            return marshal(tags, tag_fields), HTTPStatus.OK
-
-        except NotFoundException:
-            return {'message': 'table_uri {} does not exist'.format(table_uri)}, HTTPStatus.NOT_FOUND
-
-        except Exception:
-            return {'message': 'Internal server error!'}, HTTPStatus.INTERNAL_SERVER_ERROR
-
     def put(self, table_uri: str, tag: str) -> Iterable[Union[Mapping, int, None]]:
         """
         API to add a tag to existing table uri.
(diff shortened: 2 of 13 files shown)

### CODE REVIEW

File: metadata_service/__init__.py [hunk 1]
```diff
@@ -73,7 +73,7 @@ def create_app(*, config_module_class: str) -> Flask:
     api.add_resource(TableDescriptionAPI,
                      '/table/<path:table_uri>/description')
     api.add_resource(TableTagAPI,
-                     '/table/<path:table_uri>/tag',
+                     '/table/<path:table_uri>/tag/',
```
jinhyukchang: Would introducing tailing slash break backward compatibility? Is frontend using this API?

metadata_service/api/table.py
```diff
@@ -188,16 +204,31 @@ def put(self, table_uri: str, tag: str) -> Iterable[Union[Mapping, int, None]]:
...
         except NotFoundException:
             return \
-                {'message': 'The tag {} for table_uri {} '
+                {'message': 'The tag {} for table_uri {} with type {} '
                             'is not added successfully'.format(tag,
-                                                               table_uri)}, \
+                                                               table_uri,
+                                                               tag_type)}, \
+                HTTPStatus.NOT_FOUND
+        except BadgeNotInWhitelistException:
```
jinhyukchang: How about validating whitelist in the API level? So that it won't need to be implemented on multiple proxies?

metadata_service/api/table.py
```diff
@@ -208,14 +239,28 @@ def delete(self, table_uri: str, tag: str) -> Iterable[Union[Mapping, int, None]
...
         except NotFoundException:
             return \
-                {'message': 'The tag {} for table_uri {} '
+                {'message': 'The tag {} for table_uri {} with type {} '
                             'is not deleted successfully'.format(tag,
-                                                                 table_uri)}, \
+                                                                 table_uri,
+                                                                 tag_type)}, \
+                HTTPStatus.NOT_FOUND
+        except BadgeNotInWhitelistException:
```
jinhyukchang: I wonder if we need to validate whitelist on deletion. May be we can be flexible on deletion so that others can delete it when we dropped certain badges. WDYT?

(If we decided to keep this, message needs to be updated :)  s/added/deleted)

metadata_service/proxy/neo4j_proxy.py
```diff
@@ -173,7 +174,8 @@ def _exec_table_query(self, table_uri: str) -> Tuple:
         """)
 
         table_records = self._execute_cypher_query(statement=table_level_query,
-                                                   param_dict={'tbl_key': table_uri})
+                                                   param_dict={'tbl_key': table_uri,
+                                                               'tag_type': 'default'})
```
danwom: I would like to avoid making a separate call to fetch badges for each table when they could be included in the base result just like tags. This problem gets worse with search results where we'd have to make 10 queries at once (one for each search result).

Not sure about the exact syntax but could we do something like
```
  ...
  OPTIONAL MATCH (tbl)-[:TAGGED_BY]->(tag:Tag)
  ...
  collect(distinct tag:Tag{tag_type:'default'}) as tag_records,
  collect(distinct tag:Tag{tag_type:'badge'}) as badge_records,
```
We would also remove the `tag_type` parameter from `param_dict`.

danwom: Or would the syntax be 

```
...
  OPTIONAL MATCH (tbl)-[:TAGGED_BY]->(tag:Tag{tag_type: 'default'})
  OPTIONAL MATCH (tbl)-[:TAGGED_BY]->(badge:Tag{tag_type: 'badge'})
  ...
  collect(distinct tag) as tag_records,
  collect(distinct badge) as badge_records,

```

feng-tao: I think putting everything inside single call won't fly. What if we later need to add table_stats to show the table object?

metadata_service/__init__.py
feng-tao: turns out that endpoint is no use before(not sure why we create two end point for add tag), and I try to keep it the same as others.

metadata_service/api/table.py
feng-tao: sg

metadata_service/api/table.py
feng-tao: will do

metadata_service/proxy/neo4j_proxy.py
feng-tao: but if you have strong opinion this one, I could modify on Friday or early next week. WDYT @jinhyukchang

danwom: I don't see a strong reason why `badges` should be different from `tags` or why `table_stats` should be different from `col_stats`.

Calling `get_table` currently queries for tags, owners, columns, and multiple `col_stats` for each column, which can result in hundreds of `col_stat` objects. If we're concerned about adding `table_stats` or `badges`, the performance cost will be a small fraction of whatever `col_stats` costs. If we were to draw an arbitrary line about what data gets included in the base API and what gets left off, I'd argue that `col_stats` gets left off and everything else gets added in. Currently fetching table metadata takes between 100-200 ms so I don't think this is a concern yet.

On the flip side let's say we want to decomp all of these APIs. On loading the `table_details` page we would make calls to `get_table`, `get_table_tags`, `get_table_badges`, `get_table_stats`, `get_table_columns` and `get_column_stats`. I don't think this kind of micro-optimization is worth the effort.

metadata_service/proxy/neo4j_proxy.py
```diff
@@ -162,7 +163,7 @@ def _exec_table_query(self, table_uri: str) -> Tuple:
         OPTIONAL MATCH (application:Application)-[:GENERATES]->(tbl)
         OPTIONAL MATCH (tbl)-[:LAST_UPDATED_AT]->(t:Timestamp)
         OPTIONAL MATCH (owner:User)<-[:OWNER]-(tbl)
-        OPTIONAL MATCH (tbl)-[:TAGGED_BY]->(tag:Tag)
+        OPTIONAL MATCH (tbl)-[:TAGGED_BY]->(tag:Tag{tag_type: $tag_type})
```
feng-tao: @danwom , if I don't change this line, you should get both badge and tag together through the API which FE needs to change the logic to distinguish these two based on the tag_type. Does it work for you?

metadata_service/proxy/neo4j_proxy.py
jinhyukchang: Can we try query like below?:
...
OPTIONAL MATCH (tbl)-[:TAGGED_BY]->(tag:Tag{tag_type: 'default'})
OPTIONAL MATCH (tbl)-[:TAGGED_BY]->(badge:Tag{tag_type: 'badge'})
...

metadata_service/proxy/neo4j_proxy.py
feng-tao:
```
tag_fields = {
    'tag_type': fields.String,
    'tag_name': fields.String
}

table_writer_fields = {
    'application_url': fields.String(attribute='application_url'),
    'name': fields.String,
    'id': fields.String,
    'description': fields.String  # Optional
}



table_detail_fields = {
    'database': fields.String,
    'cluster': fields.String,
    'schema': fields.String,
    'table_name': fields.String(attribute='name'),
    'table_description': fields.String(attribute='description'),  # Optional
    'tags': fields.List(fields.Nested(tag_fields)),  # Can be an empty list
    # Can be an empty list
    'table_readers': fields.List(fields.Nested(table_reader_fields)),
    # Can be an empty list
    'owners': fields.List(fields.Nested(user_fields)),
    # Can be an empty list
    'columns': fields.List(fields.Nested(column_fields)),
    # Can be an empty list
    'watermarks': fields.List(fields.Nested(watermark_fields)),
    'table_writer': fields.Nested(table_writer_fields),  # Optional
    'last_updated_timestamp': fields.Integer,  # Optional
    'source': fields.Nested(source_fields),  # Optional
    'is_view': fields.Boolean  # Optional
}```

Comment (feng-tao) [reply metadata_service/proxy/neo4j_proxy.py]: this is the original payload.

Comment (feng-tao) [reply metadata_service/proxy/neo4j_proxy.py]: also @jinhyukchang

Comment (danwom) [reply metadata_service/proxy/neo4j_proxy.py]: As a last resort getting badges and tags in the `tag` field would be workable, but can't we put it in a new field `badges` using the following?

```
  OPTIONAL MATCH (tbl)-[:TAGGED_BY]->(tag:Tag{tag_type: 'default'})
  OPTIONAL MATCH (tbl)-[:TAGGED_BY]->(badge:Tag{tag_type: 'badge'})
  ...
  collect(distinct tag) as tag_records,
  collect(distinct badge) as badge_records,
```

File: metadata_service/proxy/neo4j_proxy.py (hunk 4 above)
Comment (jinhyukchang) [reply metadata_service/proxy/neo4j_proxy.py]: Oops. It's already mentioned. Shall we see how it goes? I believe we still have sometime, right?

File: metadata_service/proxy/neo4j_proxy.py (hunk 5 above)
Comment (feng-tao) [reply metadata_service/proxy/neo4j_proxy.py]: this will change table API which we need to see if it impacts atlas backend.

**Comment by feng-tao** [comment]
@jinhyukchang @danwom I updated the pr based on the discussion.

**Comment by feng-tao** [comment]
also cc @verdan for FYI we plan to add a new concept called badge which is non human editable type of tag(ingestion from endpoint) to allow tables to show sth like "golden/production ready/beta" etc.
If the table doesn't have any badge, the FE won't show, so it should behave the same as before for atlas proxy.  But once we roll out, do let us know if you are seeing any issues.

File: metadata_service/config.py [hunk 6]
```diff
@@ -34,13 +35,18 @@ class Config:
...
+    # whitelist badges
+    WHITELIST_BADGES: List[str] = []
+
 
 class LocalConfig(Config):
     DEBUG = False
     TESTING = False
     LOG_LEVEL = 'DEBUG'
-    LOCAL_HOST = '0.0.0.0'
+    LOCAL_HOST = 'amundsenneo4j-staging.lyft.net'
```
jinhyukchang: Revert this? :)

metadata_service/config.py
```diff
@@ -34,13 +35,18 @@ class Config:
...
 
     PROXY_HOST = os.environ.get('PROXY_HOST', f'bolt://{LOCAL_HOST}')
     PROXY_PORT = os.environ.get('PROXY_PORT', 7687)
     PROXY_CLIENT = PROXY_CLIENTS[os.environ.get('PROXY_CLIENT', 'NEO4J')]
+    PROXY_USER = os.environ.get('CREDENTIALS_NEO4J_USER', 'neo4j')
+    PROXY_PASSWORD = os.environ.get('CREDENTIALS_NEO4J_PASSWORD', '1234abcd!')
```
jinhyukchang: Remove these two lines as well?

metadata_service/proxy/neo4j_proxy.py
```diff
@@ -497,20 +510,49 @@ def delete_owner(self, *,
             tx.commit()
             tx.close()
 
+    @timer_with_counter
+    def get_tag(self, *, table_uri: str, tag_type: str = 'default') -> List:
```
jinhyukchang: Is there anywhere calling this method? Was not able to find it..

feng-tao: will remove it.

metadata_service/config.py
feng-tao: shxx :)

metadata_service/config.py
feng-tao: :(

verdan:
> also cc @verdan for FYI we plan to add a new concept called badge which is non human editable type of tag(ingestion from endpoint) to allow tables to show sth like "golden/production ready/beta" etc.
> If the table doesn't have any badge, the FE won't show, so it should behave the same as before for atlas proxy. But once we roll out, do let us know if you are seeing any issues.

thanks @feng-tao, this will be pretty useful to apply Ranger Policies attached to Atlas entities. 👍

PR #101, opened 2020-01-15
Add open api spec for metadata_service

dmateusp:
### Summary of Changes

Add OpenAPI (swagger) docs to metadata_service, https://github.com/lyft/amundsen/issues/166

### Tests

Brought the swagger_tests from the search_service with minor modifications (I changed the .replace() calls to cover cases with `<path:something>` inside the URL)

### Documentation

I added a reference to the API docs in the README.md

### CheckList
- [x] PR passes `make test`

### DIFF
--- a/metadata_service/api/user.py
+++ b/metadata_service/api/user.py
@@ -2,6 +2,7 @@ from http import HTTPStatus
 from typing import Iterable, Mapping, Union
 
 from flask_restful import Resource, fields, marshal
+from flasgger import swag_from
 
 from metadata_service.api.popular_tables import popular_table_fields
 from metadata_service.exception import NotFoundException
@@ -41,6 +42,7 @@ class UserDetailAPI(Resource):
 
         self.client = get_proxy_client()
 
+    @swag_from('swagger_doc/user/detail_get.yml')
     def get(self, user_id: str) -> Iterable[Union[Mapping, int, None]]:
         try:
             table = self.client.get_user_detail(user_id=user_id)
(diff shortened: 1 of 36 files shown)


dmateusp:
Looking into adding tests and documentation from search_service now

metadata_service/api/user.py
```diff
@@ -50,15 +52,15 @@ def get(self, user_id: str) -> Iterable[Union[Mapping, int, None]]:
             return {'message': 'User id {} does not exist'.format(user_id)}, HTTPStatus.NOT_FOUND
 
 
-class UserFollowAPI(Resource):
+class UserFollowsAPI(Resource):
```
feng-tao: why do we need to add new class for get instead of putting get/put etc under a same resource?

dmateusp: I think it is described in the PR that was linked in the issue you raised: https://github.com/lyft/amundsensearchlibrary/pull/60 (I was seeing the same issue in metadata service where some URLs were invalid for some methods)

> [...] However, the Swagger documentation is looking at all the endpoints and methods setup in Flask, so this bug was causing the Swagger documentation to give users bad information. 😞

> Why is this happening?

> These lines of code in the service init file are the problem. This code is adding the resource to Flask and allowing all methods to be accessed via both URLs. However, both URLs are not supported by our methods.

In summary, each class only has "one type of URL"

feng-tao: ah, forgot about that:) thanks.

PR #102, opened 2020-01-17
change default port to 5002

friendtocephalopods:
### Summary of Changes

Change the default port to 5002. Currently all wsgi try to start up with the default port, so it's easier/free to start up all 3 servers for local debugging if they each have a different port.

This is kind of speculative --  WDYT? If ya'll are like us, you've customized the port for your individual circumstances anyway, but if you're expecting the port to be 5000, this will cause some problems...

### Tests

Config/port change. No tests affected.

### Documentation

Changed port number to 5002.

### CheckList
- [x] PR passes `make test`

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@ $ pip3 install amundsen-metadata
 $ python3 metadata_service/metadata_wsgi.py
 
 -- In a different terminal, verify getting HTTP/1.0 200 OK
-$ curl -v http://localhost:5000/healthcheck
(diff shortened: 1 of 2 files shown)


feng-tao:
I think it make sense as internally we custom the port as well. But I think we need to update all the docker file(e.g https://github.com/lyft/amundsen/blob/master/docker-amundsen.yml#L50)  as it still refers to port 5000 and convert it to 5002.

Otherwise I think it will break if user use docker-compose file.

PR #105, opened 2020-01-28
use schemas, not fields

friendtocephalopods:
### Summary of Changes

Uses schemas instead of fields for serialization/deserialization. Note: *will require accompanying frontend PR to be merged/deployed simultaneously* as some field names are tweaked for consistency (e.g. schema_name -> schema, reader -> user)

### Tests

Modified tests to expect modified API shape

### Documentation

None.

### CheckList
- [x] PR passes `make test`


friendtocephalopods:
Jin, thanks for the review!

Hmm, yeah, I felt a little bad about making some breaking changes. Maybe it's a little better that the breaking changes are all batched up together? I think it would be probably more trouble than it's worth to try and version these, since IIUC folks are using this as a user-facing tool (so a little downtime is okay) and it would require significant work on the frontend to layer these changes, and then to rename things throughout the js/tsx after....if you feel strongly about it, I'd cheer you on, but upstreaming this with the significant changes to the frontend (relative to our internal fork's last rebase) already took about as much time as I can spend on this work.

Not to mention that it would be quite painful to try and version the search document model changes...
I could up the version from 1.1.x -> 1.2.0 to make it more obvious it's a breaking change if that helps? 

The good news is I did test these three PRs all together and AFAICT everything is working together nicely (popular tables, search, metadata view).

feng-tao:
thanks @jinhyukchang and @friendtocephalopods . I chatted with @friendtocephalopods and felt that this mismatched term is one of the tech debt we should fix . And given it is user facing tool and not api dependent,I felt we should be fine as long as we document it in a faq doc?

Make it API versioning seems a bit overkilled. Let me know what you think.

feng-tao:
@friendtocephalopods if you dont have time to document the deployment step, I could try to do it as well.

jinhyukchang:
@friendtocephalopods @feng-tao 
Agreed on your points. Want to bring this improvement, but at the same time minimize the impact. 

Should we do `2.0.0` for it is backward incompatible change?
https://semver.org/
`MAJOR version when you make incompatible API changes,`

Will notify the channel that this change will come.

Again, thanks @friendtocephalopods

friendtocephalopods:
Sure, I can bump the version to 2.0.0 and write a few notes about deployment. What's the preferred medium for deployment notes? a comment here? a gist? adding it to the README might be a little overkill for a transient migration, but I could!

feng-tao:
a faq doc(we could create one if not exist) in lyft/amundsen repo? Agree that putting in readme is a bit overkilled.

jornh:
I’d prefer to just put the info in https://github.com/lyft/amundsenfrontendlibrary/releases over putting it in a FAQ doc. This is transient information only relevant to those updating from versions before 2.0.0 - and they will/should naturally look at the change log.

Maybe we can even start tagging the `amundsen` umbrella ☂ repo here and put change log info up here as well? When we then tag it we should make sure to bump container versions across all relevant compose files and Kubernetes helm charts (should I raise this as a separate issue?). 

That’s not to say a FAQ document isn’t valuable for a lot of other small cross-repo cutting info or insights you’re not sure where to put.

👏 to @jinhyukchang for raising the `2.0.0` suggestion and the @here announcement + pin on slack. That’s really good style! 
I’d really like if `2.0.0` could get out in about a week (next Friday?) including the other breaking change proposed in  #109, and be done with it. With the Slack announcement out that should strike a balance with people being warned and with light and swift forward movement of Amundsen.

jinhyukchang:
@friendtocephalopods Tested all services and looks good. Could you resolve conflicts on this PR and then we are good to go.

requirements.txt
```diff
@@ -41,7 +41,7 @@ pytest-mock==1.1
 typing==3.6.4
 
 
-amundsen-common==0.1.3
+amundsen-common==0.1.6
```
jinhyukchang: Could you change this to `0.2.2`?

PR #110, opened 2020-02-04
Switch from user_id to id in get_user method

alran:
### Summary of Changes

A follow up to my previous PR, which adds a BaseAPI with some generic logic. That logic expects "id" as an argument for the "get" method. This PR replaces user_id with id. 

### Tests

I updated the user API test to use `assert_called_once_with` so that this bug can be caught in the future via automated testing.

### CheckList
- [x] PR passes `make test`

### DIFF
--- a/metadata_service/proxy/atlas_proxy.py
+++ b/metadata_service/proxy/atlas_proxy.py
@@ -282,7 +282,7 @@ class AtlasProxy(BaseProxy):
             )
         return sorted(columns, key=lambda item: item.sort_order)
 
-    def get_user(self, *, user_id: str) -> Union[UserEntity, None]:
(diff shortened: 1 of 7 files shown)


alran:
Adding some notes here (cc @feng-tao):
- the front end expects "user_id"but passes it through as part of the route like 
```
 url = '{0}{1}/{2}'.format(app.config['METADATASERVICE_BASE'], USER_ENDPOINT, user_id)
```
- on the metadata side, the route is named like
```
api.add_resource(UserDetailAPI, '/user', '/user/<path:id>')
```
- There are additional user endpoints like Follow/Own/Read which expect user_id. I would argue that all of these should move towards "id" so that we can switch to using the BaseAPI

PR #99, opened 2020-01-09
Clean up user endpoints and add BaseAPI

alran:
### Summary of Changes

This PR adds tests for user endpoints that were previously untested. It also adds a BaseAPI and uses this BaseAPI in the UserDetailAPI. 

### Tests

I added tests for the user endpoints.

### CheckList
Make sure you have checked **all** steps below to ensure a timely review.
- [x] PR title addresses the issue accurately and concisely. Example: "Updates the version of Flask to v1.0.2"
    - In case you are adding a dependency, check if the license complies with the [ASF 3rd Party License Policy](https://www.apache.org/legal/resolved.html#category-x).
- [x] PR includes a summary of changes. 
- [x] PR adds unit tests, updates existing unit tests, __OR__ documents why no test additions or modifications are needed.
- [x] In case of new functionality, my PR adds documentation that describes how to use it.
    - All the public functions and the classes in the PR contain docstrings that explain what it does
- [x] PR passes `make test`

PR #113, opened 2020-02-06
Reverting user follows API from using PopularTableSchema

jinhyukchang:
### Summary of Changes

Recent [PR](https://github.com/lyft/amundsenmetadatalibrary/pull/99) brought backward incompatible change to "user follows" API and this PR is to revert the change.

Also, uses amundsen-common==0.1.3rc0 for 0.1.3 is for Python>=3.7 and it is not compatible with other micro-services.

This is a short term fix to fix broken master branch. 

### Tests

Integration test done.

### Documentation

N/A

### CheckList
- [x] PR passes `make test`

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -41,7 +41,7 @@ pytest-mock==1.1
 typing==3.6.4
 
 
-amundsen-common==0.1.7rc1
+amundsen-common==0.1.3rc0
 flasgger==0.9.3
 Flask-RESTful==0.3.6
 Flask==1.0.2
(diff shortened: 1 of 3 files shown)


requirements.txt
```diff
@@ -41,7 +41,7 @@ pytest-mock==1.1
 typing==3.6.4
 
 
-amundsen-common==0.1.3
+amundsen-common==0.1.3rc0
```
feng-tao: should it be rc0?

jinhyukchang: Yes, 0.1.3 requires Python 3.7 where metadata service is not. `rc0` is the one with Python 3.6.

I think we can quickly get out of this situation when metadata service adopts amundsen-common==0.1.7 just released with Python 3.6

PR #133, opened 2020-05-04
Fix Bookmarks after latest changes (AtlasProxy)

mgorsk1:
### Summary of Changes

With this PR:

- Bookmarks feature has been brought back to life
- Issue with retrieving other users bookmarks was fixed - when user A bookmarked user's B table, user B could also see it in his `My Bookmarks` section (assuming that db name is users id)
- Retrieval of deleted Bookmarks has been turned off
- New methods are added (create_user, user_exists) that can be used to add users to atlas automatically

Please note that it requires amundsenatlastypes in **1.1.0** version, where `Bookmark` entity was introduced.

### Tests

Aligned tests to new entity definitions.

### Documentation

### CheckList
- [x] PR passes `make test`

### DIFF
--- a/metadata_service/proxy/atlas_proxy.py
+++ b/metadata_service/proxy/atlas_proxy.py
@@ -111,6 +111,28 @@ class AtlasProxy(BaseProxy):
         result = pattern.match(table_uri)
         return result.groupdict() if result else dict()
 
+    def _parse_reader_qn(self, reader_qn: str) -> Dict:
+        """
+        Parse reader qualifiedName and extract the info
+        :param reader_qn:
+        :return: Dictionary object containing following information:
+        cluster: cluster information
+        db: Database name
+        name: Table name
+        """
+        pattern = re.compile(r"""
+        ^(?P<db>[^.]*)
+        \.
+        (?P<table>[^.]*)
+        \.
+        (?P<user_id>[^.]*)\.reader
+        \@
+        (?P<cluster>.*)
+        $
+        """, re.X)
+        result = pattern.match(reader_qn)
+        return result.groupdict() if result else dict()
+
     def _parse_bookmark_qn(self, bookmark_qn: str) -> Dict:
         """
         Parse bookmark qualifiedName and extract the info
@@ -201,10 +223,7 @@ class AtlasProxy(BaseProxy):
         """
         user_entity = self._driver.entity_unique_attribute(self.USER_TYPE, qualifiedName=user_id)
 
-        if not user_entity.entity:
-            return False
-        else:
-            return True
+        return bool(user_entity.entity)
 
     def create_user(self, user_id: str) -> None:
         """
--- a/metadata_service/proxy/atlas_proxy.py
+++ b/metadata_service/proxy/atlas_proxy.py
@@ -42,7 +42,7 @@ class AtlasProxy(BaseProxy):
     BOOKMARK_TYPE = 'Bookmark'
     USER_TYPE = 'User'
     QN_KEY = 'qualifiedName'
-    BOOMARK_ACTIVE_KEY = 'active'
+    BOOKMARK_ACTIVE_KEY = 'active'
     GUID_KEY = 'guid'
     ATTRS_KEY = 'attributes'
     REL_ATTRS_KEY = 'relationshipAttributes'
@@ -184,7 +184,7 @@ class AtlasProxy(BaseProxy):
             'entity': {
                 'typeName': self.BOOKMARK_TYPE,
                 'attributes': {'qualifiedName': bookmark_qn,
-                               self.BOOMARK_ACTIVE_KEY: True,
+                               self.BOOKMARK_ACTIVE_KEY: True,
                                'entityUri': table_uri,
                                'user': {'guid': user_guid},
                                'entity': {'guid': entity.entity[self.GUID_KEY]}}
@@ -569,7 +569,7 @@ class AtlasProxy(BaseProxy):
                         'attributeValue': f'.{user_email}.bookmark'
                     },
                     {
-                        'attributeName': self.BOOMARK_ACTIVE_KEY,
+                        'attributeName': self.BOOKMARK_ACTIVE_KEY,
                         'operator': 'eq',
                         'attributeValue': 'true'
                     }
@@ -614,7 +614,7 @@ class AtlasProxy(BaseProxy):
                                     relation_type: UserResourceRel) -> None:
 
         entity = self._get_bookmark_entity(entity_uri=table_uri, user_id=user_email)
-        entity.entity[self.ATTRS_KEY][self.BOOMARK_ACTIVE_KEY] = True
+        entity.entity[self.ATTRS_KEY][self.BOOKMARK_ACTIVE_KEY] = True
         entity.update()
 
     def delete_resource_relation_by_user(self, *,
@@ -634,7 +634,7 @@ class AtlasProxy(BaseProxy):
                                        user_email: str,
                                        relation_type: UserResourceRel) -> None:
         entity = self._get_bookmark_entity(entity_uri=table_uri, user_id=user_email)
-        entity.entity[self.ATTRS_KEY][self.BOOMARK_ACTIVE_KEY] = False
+        entity.entity[self.ATTRS_KEY][self.BOOKMARK_ACTIVE_KEY] = False
         entity.update()
 
     def get_dashboard(self,
--- a/metadata_service/proxy/atlas_proxy.py
+++ b/metadata_service/proxy/atlas_proxy.py
@@ -215,34 +215,6 @@ class AtlasProxy(BaseProxy):
 
         self._driver.entity_post.create(data=bookmark_entity)
 
-    def user_exists(self, user_id: str) -> bool:
-        """
-        Verifies if given user exists in Atlas.
-        :param user_id: User's id
-        :return: True if user exists, otherwise False
-        """
-        user_entity = self._driver.entity_unique_attribute(self.USER_TYPE, qualifiedName=user_id)
-
-        return bool(user_entity.entity)
-
-    def create_user(self, user_id: str) -> None:
-        """
-        Creates a User entity.
-        :param user_id: User's id
-        :return:
-        """
-        user_entity = {
-            'entity': {
-                'typeName': self.USER_TYPE,
-                'attributes': {'qualifiedName': user_id}
-            }
-        }
-
-        try:
-            self._driver.entity_post.create(data=user_entity)
-        except Exception as ex:
-            LOGGER.error('Error creating user.', exc_info=True)
-
     def _get_bookmark_entity(self, entity_uri: str, user_id: str) -> EntityUniqueAttribute:
         """
         Fetch a Bookmark entity from parsing table uri and user id.
--- a/metadata_service/proxy/atlas_proxy.py
+++ b/metadata_service/proxy/atlas_proxy.py
@@ -39,9 +39,10 @@ class AtlasProxy(BaseProxy):
     TABLE_ENTITY = app.config['ATLAS_TABLE_ENTITY']
     DB_ATTRIBUTE = app.config['ATLAS_DB_ATTRIBUTE']
     STATISTICS_FORMAT_SPEC = app.config['STATISTICS_FORMAT_SPEC']
-    READER_TYPE = 'Reader'
+    BOOKMARK_TYPE = 'Bookmark'
+    USER_TYPE = 'User'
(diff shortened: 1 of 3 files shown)


mgorsk1:
cc @verdan

verdan:
@mgorsk1 In my opinion we should revisit the bookmarks feature completely. I don't like the approach we had taken for this before. Reader entity has nothing to do with the Bookmarks logically, but we are storing the bookmarks boolean in the Reader entity. 
I'd revisit the entity definition, and simply create a list of entities under the User entity, and store a list of entities (could be table, dashboard etc.) under that.

mgorsk1:
I was also thinking about refactoring it but something less invasive. In previous design, reader was also meant for popularity score, that's why we keep `count` property there. And I would get rid of that.

As for the `Reader` entity, this kind of bridge between `User` and entity that he bookmarked has it's benefits:

- once Reader is created, you don't really have to 'touch' User/Table entities to manage bookmarks (just switching the flag)
- ability to store `entityUri` in dedicated object
- changing bookmark doesn't require acquiring whole `User` object with it's relationships (which can be slower)

verdan:
-  The ‘count’ property under the Reader is to support the frequent users. 
-  I see no reason to store the ‘entityUri’ anwhere in the system since we can already create one in runtime. 


Storing Bookmarks under User would be simple in many ways:
-  User will have all information about its logical relationships. 
-  Only one query to Users to get all relationships of the object, which is indexed properly based on qualified name
-  Multiple types of Bookmarks support would be easy i.e., Table, Dashboard etc. 
-  No need to create the Reader entity and complicated operations everytime if it’s not there.

mgorsk1:
- In regards to `count` - I don't think it provides any value with implementation we have right now (what would count 5 mean ?).
- I understand `entityUri` as a pseudo-serialization effort that enables us to become entity agnostic on metadata side. Rendering it in metadata during runtime would mean maintaining separate logic for each supported entity.

I understand the benefits and drawbacks of both approaches, my concern is about making such large change out of the blue - especially that we already went through two entities refactoring (last one in March) and it never came up then.

Adding @dwarszawski to the discussion as he was involved in second iteration of amundsenatlastypes refactoring in March.

verdan:
> * In regards to `count` - I don't think it provides any value with implementation we have right now - we don't have process updating it + I wouldn't know how to interpret value 5 if it was there.
> * I understand `entityUri` as a pseudo-serialization effort that enables us to become entity agnostic on metadata side. Rendering it in metadata during runtime would mean maintaining separate logic for each supported entity.
> * We still have relationship `User` <> `Reader` so User will have all information about its logical relationships.
> 
> I think it is a good idea to refactor `get_table_by_user_relation` to acquire bookmakrs the way you propose (so query for User entity and get all relationships), the only drawback I see is it would require two hits to Atlas instead of one.
> 
> My concern is about making this PR any more complicated - especially that we already went through two entities refactoring efforts (last one in March) and it never came up then. And with this fix bookmarks will work without any limitations.
> 
> Adding @dwarszawski to the discussion as he was involved in second iteration of amundsenatlastypes as well.

So the count in Reader entity is used to sort the Frequent Users.

mgorsk1:
Ok, so when we have process to update it - absolutely. Then frequent users would be those in top 5 (for example) after sorting by count. Still I am not sure how this collides with bookmarks feature.

verdan:
> Ok, so when we have process to update it - absolutely. Then frequent users would be those in top 5 (for example) after sorting by count. Still I am not sure how this collides with bookmarks feature. We will still need Reader to make `Frequent Users`, and not using it for bookmarks would mean twice as much relationships for User and Table (one for Reader, one for Bookmark)

Sorry I think there is a misunderstanding. It has nothing to do with the Bookmarks, just replying to your first message `In previous design, Reader was also meant for popularity score, that's why we keep count property there. And I would get rid of that.`

metadata_service/proxy/atlas_proxy.py
```diff
@@ -168,57 +171,93 @@ def _get_user_entity(self, user_id: str) -> EntityUniqueAttribute:
...
+            }
         }
 
-        self._driver.entity_post.create(data=reader_entity)
-        table_entity.entity[self.ATTRS_KEY]['readers'].append(reader_entity)
-        table_entity.update()
+        self._driver.entity_post.create(data=bookmark_entity)
 
-    def _get_reader_entity(self, table_uri: str, user_id: str) -> EntityUniqueAttribute:
+    def user_exists(self, user_id: str) -> bool:
```
dwarszawski, line 196: is that used somewhere ?

metadata_service/proxy/atlas_proxy.py
```diff
@@ -168,57 +171,93 @@ def _get_user_entity(self, user_id: str) -> EntityUniqueAttribute:
...
 
-    def _create_reader(self, table_entity: EntityUniqueAttribute, user_guid: str, reader_qn: str) -> None:
+    def _create_bookmark(self, entity: EntityUniqueAttribute, user_guid: str, bookmark_qn: str,
+                         table_uri: str) -> None:
         """
-        Creates a reader entity for a specific user and table uri.
+        Creates a bookmark entity for a specific user and table uri.
         :param user_guid: User's guid
-        :param reader_qn: Reader qualifiedName
+        :param bookmark_qn: Bookmark qualifiedName
```
dwarszawski, line 179: is the concept of Reader still applicable or that should be removed from amundsenatlastypes ?

metadata_service/proxy/atlas_proxy.py
```diff
@@ -168,57 +171,93 @@ def _get_user_entity(self, user_id: str) -> EntityUniqueAttribute:
...
-            'attributes': {'qualifiedName': reader_qn,
-                           'isFollowing': True,
-                           'count': 0,
-                           'user': {'guid': user_guid}}
+
+        bookmark_entity = {
+            'entity': {
+                'typeName': self.BOOKMARK_TYPE,
+                'attributes': {'qualifiedName': bookmark_qn,
+                               self.BOOKMARK_ACTIVE_KEY: True,
```
dwarszawski, line 187: what is the reason of having this `active`flag. Is that not reflected by simply deleting the relationship and bookmark entity?

metadata_service/proxy/atlas_proxy.py
mgorsk1: this is used in our get_user_details function when acquiring ldap data

metadata_service/proxy/atlas_proxy.py
mgorsk1: it's still applicable and should be used for frequent-users functionality

metadata_service/proxy/atlas_proxy.py
mgorsk1: we don't really rely on relationship while searching. I could delete the Bookmark entity when user `unsubscribes` but when when he `subscribes` again it would produce new entity object with different guid. And making/deleting bookmark might be frequently done in regards to the same table. That's why instead of deleting and recreating (which produces new object in storage) every time, I rely on active flag.

dwarszawski: That can lead to explosion of not-referred entities. Probably, that is already applied in some way with `exludeDeletedEntities` filter in Atlas but that is not dedicated to recover entities and kind of automatically cleaned by Atlas. I would rather delete not used entities as that can introduce additional level of complexity while working with this type. Let's keep things as simple as possible.

mgorsk1: Are you certain that there is a process automatically cleaning inactive entities ? I have never seen it working like that to be honest, inactive entities just stay in Atlas.

Either approach `could` lead to explosion of entities, the question is which one will do more harm:

-  Creating bookmark means not only new entity with the same qualified name but different guid. If we were to drop it completely, every new `create` operation on the same table would be 3 new objects (bookmark itself and two relationship definitions)
-  Switching active flag `true/false` means just one set of 3 entities regardless of number of times it's done.

dwarszawski: Ok, it's not automatic but at least there is dedicated purge API(/admin/purge) to delete soft deleted entities. So approach with additional flag is inconsistent with general atlas entity life cycle.

mgorsk1: I agree purging (if done periodically) could resolve this issue but leaving a room for a malicious user to exploit that is also something we should take into consideration. After all, bookmarks feature in general does not adhere to general atlas entity life cycle and since it's exposed to be edited by user, such risks should be mitigated.

dwarszawski: Every entity should undego same life cycle while considering general view. Additionally should be decoupled from business requirements. Any exception to this introduces confusions.

mgorsk1: Entity in Atlas by design is a byproduct of users interaction with data, not something he explicitly creates. So not every entity undergoes the same life cycle. Hence, I do not agree that it fits `general view`. And it's not a business requirement but technical one.

I am really surprised this is something that was not raised during entity design and approval.

metadata_service/proxy/atlas_proxy.py
```diff
@@ -39,9 +39,10 @@ class AtlasProxy(BaseProxy):
     TABLE_ENTITY = app.config['ATLAS_TABLE_ENTITY']
     DB_ATTRIBUTE = app.config['ATLAS_DB_ATTRIBUTE']
     STATISTICS_FORMAT_SPEC = app.config['STATISTICS_FORMAT_SPEC']
-    READER_TYPE = 'Reader'
+    BOOKMARK_TYPE = 'Bookmark'
+    USER_TYPE = 'User'
     QN_KEY = 'qualifiedName'
-    BKMARKS_KEY = 'isFollowing'
+    BOOKMARK_ACTIVE_KEY = 'active'
```
verdan, line 45: Do you need to make this a constant ? not adding any extra value, just making things complex to read.

metadata_service/proxy/atlas_proxy.py
```diff
@@ -110,10 +111,10 @@ def _extract_info_from_uri(self, *, table_uri: str) -> Dict:
         result = pattern.match(table_uri)
         return result.groupdict() if result else dict()
 
-    def _parse_reader_qn(self, reader_qn: str) -> Dict:
+    def _parse_bookmark_qn(self, bookmark_qn: str) -> Dict:
```
verdan, line 114: I will not remove this `_parse_reader_qn` function, and create another for bookmarks, since it will be used for the user profile and frequent users.

metadata_service/proxy/atlas_proxy.py
```diff
@@ -168,57 +171,93 @@ def _get_user_entity(self, user_id: str) -> EntityUniqueAttribute:
...
+        """
+        Verifies if given user exists in Atlas.
+        :param user_id: User's id
+        :return: True if user exists, otherwise False
         """
-        Fetch a Reader entity from parsing table uri and user id.
-        If Reader is not present, create one for the user.
+        user_entity = self._driver.entity_unique_attribute(self.USER_TYPE, qualifiedName=user_id)
+
+        if not user_entity.entity:
```
verdan, line 204: you mean 😉 
```
return bool(user_entity.entity)
```

metadata_service/proxy/atlas_proxy.py
```diff
@@ -168,57 +171,93 @@ def _get_user_entity(self, user_id: str) -> EntityUniqueAttribute:
...
+            return False
+        else:
+            return True
+
+    def create_user(self, user_id: str) -> None:
```
verdan, line 209: what does this function do ? where it is being used ?

metadata_service/proxy/atlas_proxy.py
mgorsk1: good catch, revereted deleting of this one

metadata_service/proxy/atlas_proxy.py
mgorsk1: I added it as a helper method and am using in in our `get_user_details` function in metadata config

metadata_service/proxy/atlas_proxy.py
mgorsk1: yep. we can do it like that

metadata_service/proxy/atlas_proxy.py
mgorsk1: it's used in 4 different places throughout this class, so for consistency it made sense to define it as constant

verdan:
Looks good to me 👍 
Thank you @mgorsk1

feng-tao:
thanks everyone!

PR #134, opened 2020-05-04
Hide inactive columns in table view (AtlasProxy)

mgorsk1:
### Summary of Changes

When listing columns in tables view there is no check if the column is still active. This might lead to showing inaccurate results.

### Tests

Aligned, extended & run to completion.

### Documentation

### CheckList
- [x] PR passes `make test`

### DIFF
--- a/metadata_service/proxy/atlas_proxy.py
+++ b/metadata_service/proxy/atlas_proxy.py
@@ -283,6 +283,11 @@ class AtlasProxy(BaseProxy):
         """
         columns = list()
         for column in entity.entity[self.REL_ATTRS_KEY].get('columns') or list():
+            column_status = column.get('entityStatus', 'inactive').lower()
+
+            if column_status != 'active':
+                continue
+
             col_entity = entity.referredEntities[column[self.GUID_KEY]]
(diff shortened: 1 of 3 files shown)


metadata_service/proxy/atlas_proxy.py
```diff
@@ -283,6 +283,11 @@ def _serialize_columns(self, *, entity: EntityUniqueAttribute) -> \
         """
         columns = list()
         for column in entity.entity[self.REL_ATTRS_KEY].get('columns') or list():
+            column_status = column.get('entityStatus', 'inactive').lower()
+
+            if column_status != 'active':
```
feng-tao, line 288: will atlas do any stale column metadata pruning besides just adding the attribute? For example, we clean up the stale metadata in databuilder in https://github.com/lyft/amundsendatabuilder#removing-stale-data-in-neo4j----neo4jstalenessremovaltask .

mgorsk1: I think there is an endpoint to do it manually but there is no periodic flush. In majority of our searches we can just filter out inactive entities query-time, but in this it's not really an option.

PR #137, opened 2020-05-20
List Dashboards using Table API

jinhyukchang:
### Summary of Changes

Added new API that lists Dashboards using certain Table

### Tests

Unit test and integration test

### Documentation

N/A

### CheckList
- [x] PR passes `make test`


metadata_service/api/table.py
```diff
@@ -152,3 +154,17 @@ def delete(self, id: str, tag: str) -> Iterable[Union[Mapping, int, None]]:
                                        resource_type=ResourceType.Table,
                                        tag=tag,
                                        tag_type=tag_type)
+
+
+class TableDashboardAPI(BaseAPI):
```
feng-tao, line 159: maybe a comment for what this API is for?

metadata_service/entity/dashboard_summaries.py
```diff
@@ -0,0 +1,16 @@
+from typing import List
+
+import attr
+from amundsen_common.models.dashboard import DashboardSummary
+from marshmallow_annotations.ext.attrs import AttrsSchema
+
+
+@attr.s(auto_attribs=True, kw_only=True)
+class DashboardSummaries:
```
feng-tao, line 9: should it be plural for the name or singular as we did with common repo?

metadata_service/proxy/neo4j_proxy.py
```diff
@@ -1177,3 +1177,46 @@ def put_dashboard_description(self, *,
...
+        WITH c, dg, d, description, last_success_exec, sum(read.read_count) as recent_view_count
+        RETURN
+        d.key as uri,
+        c.name as cluster,
+        dg.name as group_name,
+        dg.dashboard_group_url as group_url,
+        d.name as name,
+        d.dashboard_url as url,
+        description.description as description,
+        split(d.key, '_')[0] as product,
```
feng-tao, line 1210: I forgot, but do we have product in any node/ attribute of the dashboard node? any reason we need to do a split to get the product instead of directly fetch the node?

metadata_service/entity/dashboard_summaries.py
jinhyukchang: This model is actually just plural of amundsen-common's `DashboardSummary`. So I guess just stay with "DashboardSummaries" as DashboardSummary is already taken?

metadata_service/proxy/neo4j_proxy.py
jinhyukchang: Yeah, we actually don't have it currently.

May be we can have product node just like Database node later on?

metadata_service/entity/dashboard_summaries.py
feng-tao: maybe we could `from amundsen_common.models.dashboard import DashboardSummary as summary` ?

feng-tao: just mostly felt we don't use plural anywhere else in the code base so it is not that consistent.

metadata_service/proxy/neo4j_proxy.py
feng-tao: yeah, having a product node seems better.

metadata_service/entity/dashboard_summaries.py
jinhyukchang: Sounds good :)

PR #139, opened 2020-05-27
Compatibility with Neo4J 4.0.4 (4.x in general should work)

davcamer:
### Summary of Changes

* Allow configuration to turn SSL/TLS encryption on or off for the proxy connection. Necessary to keep the Quick Start working smoothly.
* Upgrade the Neo4j driver, so that it can make insecure bolt connections. The earlier driver did not even have this option.

This works together with the changes in this PR: https://github.com/lyft/amundsendatabuilder/pull/270

### Tests

I did not see any tests that cover this configuration, or that need to change as a result. 

### Documentation

Other than the docstring for the attribute in the config class, I did not add other documentation. I'm happy to, of course, if there's somewhere I should.

### CheckList
- [x] PR passes `make test`


feng-tao:
thanks @davcamer , I wonder whether you have a chance to test the change on neo4j 3.x as well and see if both 3.x and 4.x are compatible with this change?

davcamer:
@feng-tao I have done one manual round of testing with Neo4j 3.3.0. I did this by changing the image tag in `docker-amundsen-local.yml`, and also changing this newly introduced `PROXY_ENCRYPTED` property. It works provided the client is set with the correct on/off encryption setting.

Are there "integration" style tests anywhere that should be expanded to test against both Neo4j versions?

davcamer:
I have local changes to the docker compose files in https://github.com/lyft/amundsen to use neo4j 4.0.4. It seems it would be better to have a specific tagged docker image of this project, before setting up a PR for those changes, if they are desired.

feng-tao:
thanks, will take a look, currently we don't have good integration test to test different version of the component unfortunately.

metadata_service/config.py
```diff
@@ -29,6 +31,8 @@ class Config:
 
     PROXY_USER = os.environ.get('CREDENTIALS_PROXY_USER', 'neo4j')
     PROXY_PASSWORD = os.environ.get('CREDENTIALS_PROXY_PASSWORD', 'test')
+    PROXY_ENCRYPTED = True
```
jinhyukchang, line 34: Was this off by default? If then, let's keep default `False`.

davcamer: This was previously on by default, and is on by default [in the driver](https://neo4j.com/docs/api/python-driver/1.7/driver.html#encrypted). Double checking the docs reminded me that I also want to be able to require certificate validation, which is the [trust](https://neo4j.com/docs/api/python-driver/1.7/driver.html#trust) parameter on the driver. My goal is to have full control over the TLS/SSL configuration. Should I include exposing the trust parameter in this PR?

jinhyukchang: > This was previously on by default, and is on by default [in the driver](https://neo4j.com/docs/api/python-driver/1.7/driver.html#encrypted). Double checking the docs reminded me that I also want to be able to require certificate validation, which is the [trust](https://neo4j.com/docs/api/python-driver/1.7/driver.html#trust) parameter on the driver. My goal is to have full control over the TLS/SSL configuration. Should I include exposing the trust parameter in this PR?

@davcamer I see. It seems `encrypted` is `True` when TLS is enabled from their [doc](https://neo4j.com/docs/api/python-driver/1.7/driver.html#encrypted). I wonder what would happen when `TLS` is not enabled, but forcing `encrypted` to be `True`. Not sure if it's going to fail.

davcamer: I believe by "Defaults to True if TLS is available," they mean that it is supported by the python run-time that is executing the code. From browsing their code, this is the section that determines if it is available: https://github.com/neo4j-drivers/neobolt/blob/1.7/neobolt/compat/ssl.py#L23-L35

I don't see exactly where the conflicting configuration would be caught, so this could require some Amundsen users to update their config if they:
* Are running a client in a python runtime that doesn't support SSL.
* Configured their older Neo4j to accept _unencrypted_ bolt connections. Neo4j expects encrypted connections, even on older versions.

To preserve the existing behavior, I guess I could change this to set `encrypted = None` by default? This piece of code looks to expect a None, and automatically determine in that case:
https://github.com/neo4j/neo4j-python-driver/blob/1.7.6/neo4j/__init__.py#L226-L227

jinhyukchang: @davcamer Sounds good to me. Let me know once you make a change. I tested out your branch and it works fine with me.

davcamer: I've made this change now. `PROXY_VALIDATE_SSL` is the option. It defaults to trust all certificates (no validation), since that was the previous behavior.

metadata_service/config.py
```diff
@@ -29,6 +32,10 @@ class Config:
...
```
JacobSMoller, line 35: @davcamer  little late to the party. But is there a reason this is not controlled with env vars ?

PR #144, opened 2020-06-22
fix: Returns the empty dictionary to gracefully fail

verdan:
### Summary of Changes

Passing the function causes 404, and thus crashing the frontend. This will make sure not to crash the frontend and simply returns 0 resources. 

Second fix is when there is no position attribute available for table, we can't apply sort function on those tables. hence making it 9999 to simply put that table at the end. 

### Tests

NA

### Documentation

NA

### CheckList
- [x] PR passes `make test`


feng-tao:
thanks @verdan for the quick fix.

PR #152, opened 2020-07-10
feat: Add programmatic descriptions parser for [AtlasProxy]

mgorsk1:
### Summary of Changes

This MR introduces parsing of programmatic descriptions using **parameters** field of Table entity in Atlas Proxy. The parameters field is a map of String -> Any and seems like a perfect fit for this usecase.

Moreover, this property can be set for example on hive_table entity programatically with spark sql:

`ALTER [TABLE|VIEW] table_name SET TBLPROPERTIES (key1=val1, key2=val2, ...)` - Atlas hive-hook will propagate such action in Hive Metastore to Atlas metadata

All of those key1, key2 etc then become programmatic description entries.

The idea is also to have a filter to remove unwanted properties (like spark technical ones that appear in parameters after creating table with spark). This could also be of use for other proxies.

### Tests

Tested for presence and filtering of programmatic desriptions.

### Documentation

### CheckList
- [x] PR passes `make test`

### DIFF
--- a/metadata_service/proxy/atlas_proxy.py
+++ b/metadata_service/proxy/atlas_proxy.py
@@ -4,7 +4,7 @@ from random import randint
 from typing import Any, Dict, List, Union, Optional
 
 from amundsen_common.models.popular_table import PopularTable
-from amundsen_common.models.table import Column, Statistics, Table, Tag, User, Reader
+from amundsen_common.models.table import Column, Statistics, Table, Tag, User, Reader, ProgrammaticDescription
 from amundsen_common.models.user import User as UserEntity
 from amundsen_common.models.dashboard import DashboardSummary
 from atlasclient.client import Atlas
@@ -363,6 +363,8 @@ class AtlasProxy(BaseProxy):
         try:
             attrs = table_details[self.ATTRS_KEY]
 
+            programmatic_descriptions = self._get_programmatic_descriptions(attrs.get('parameters'))
+
             table_qn = parse_table_qualified_name(
                 qualified_name=attrs.get(self.QN_KEY)
             )
@@ -389,7 +391,8 @@ class AtlasProxy(BaseProxy):
                 owners=[User(email=attrs.get('owner'))],
                 columns=columns,
                 table_readers=self._get_readers(attrs.get(self.QN_KEY)),
-                last_updated_timestamp=self._parse_date(table_details.get('updateTime')))
+                last_updated_timestamp=self._parse_date(table_details.get('updateTime')),
+                programmatic_descriptions=programmatic_descriptions)
 
             return table
         except KeyError as ex:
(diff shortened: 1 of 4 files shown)


mgorsk1:
cc @verdan

feng-tao:
cc @verdan

metadata_service/proxy/atlas_proxy.py
```diff
@@ -736,6 +739,31 @@ def _get_readers(self, qualified_name: str, top: Optional[int] = 15) -> List[Rea
...
         return results
 
+    def _get_programmatic_descriptions(self, parameters: dict) -> List[ProgrammaticDescription]:
+        _programmatic_descriptions = {}
+        programmatic_descriptions: List[ProgrammaticDescription] = []
+
+        for source, text in parameters.items():
+            use_parameter = True
+
+            for f in app.config['PROGRAMMATIC_DESCRIPTIONS_EXCLUDE_FILTERS']:
```
verdan, line 749: nit: `for filter in app.config['PROGRAMMATIC_DESCRIPTIONS_EXCLUDE_FILTERS']:`

metadata_service/proxy/atlas_proxy.py
```diff
@@ -736,6 +739,31 @@ def _get_readers(self, qualified_name: str, top: Optional[int] = 15) -> List[Rea
 
         return results
 
+    def _get_programmatic_descriptions(self, parameters: dict) -> List[ProgrammaticDescription]:
+        _programmatic_descriptions = {}
+        programmatic_descriptions: List[ProgrammaticDescription] = []
+
+        for source, text in parameters.items():
```
verdan, line 746: Two things here:
- Check if `app.config['PROGRAMMATIC_DESCRIPTIONS_EXCLUDE_FILTERS']` exists, before the loop, otherwise you are looping every time. 
- You don't need two separate variables to do this operation. i.e., within the same first loop, create the `programmatic_descriptions.append(ProgrammaticDescription(source=source, text=text))` and it will work. 

And if you want to sort then, simply apply sorting afterwards.

mgorsk1: done!

metadata_service/proxy/atlas_proxy.py
mgorsk1: `filter` shadows built-in name, I've changed to `regex_filter`

pPanda-beta:
@mgorsk1 
I see there is a particular part which breaks the camel  case field into space separated words. Looks like an UI logic. Any particular reason to make it mandatory in the metadata service instead of frontend?
 
### Source: 
https://github.com/amundsen-io/amundsenmetadatalibrary/blob/8c1b9989854fe08407af5d7da24c8e49042f9361/metadata_service/proxy/atlas_proxy.py#L961

### My previous comment on test case: 
https://github.com/amundsen-io/amundsenmetadatalibrary/commit/e8b46a48506004afe63ebbc7b721dde5a59fa07f#r42358611

PR #157, opened 2020-07-30
fix: removing OidcConfig file and making statsd configurable through envrionment variable

ayushchauhan0811:
### Summary of Changes

Removing redundant OidcConfig file from metadata service as discussed in this issue https://github.com/lyft/amundsen/issues/580

Also making statsd configurable through the environment varible `IS_STATSD_ON`


feng-tao:
cc @verdan

PR #198, opened 2020-09-09
feat: Add Watermark entries for partitioned tables in AtlasProxy

mgorsk1:
Signed-off-by: mgorsk1 <gorskimariusz13@gmail.com>

### Summary of Changes

When a Table entry contains `partitions` entries, we should verify if these partitions are time based. And if so - Watermark objects (low/high) should be created and added to Table object. 

The assumption is that `name` of partition contains only partition value and we can parse this value to match one of the expected date formats (configured with WATERMARK_DATE_FORMATS option). Another assumption is that date format is consistent throughout all partitions of table.

### Tests

Added tests with `partition` entities

### CheckList

- [x] PR passes `make test`

### DIFF
--- a/metadata_service/proxy/atlas_proxy.py
+++ b/metadata_service/proxy/atlas_proxy.py
@@ -1,15 +1,16 @@
 # Copyright Contributors to the Amundsen project.
 # SPDX-License-Identifier: Apache-2.0
 
+import datetime
 import logging
 import re
 from random import randint
-from typing import Any, Dict, List, Union, Optional
+from typing import Any, Dict, List, Union, Optional, Tuple
 
 from amundsen_common.models.dashboard import DashboardSummary
 from amundsen_common.models.popular_table import PopularTable
 from amundsen_common.models.table import Column, Stat, Table, Tag, User, Reader, \
-    ProgrammaticDescription, ResourceReport
+    ProgrammaticDescription, ResourceReport, Watermark
 from amundsen_common.models.user import User as UserEntity
 from atlasclient.client import Atlas
 from atlasclient.exceptions import BadRequest, Conflict, NotFound
@@ -450,7 +451,8 @@ class AtlasProxy(BaseProxy):
 
             reports_guids = [report.get("guid") for report in attrs.get("reports") or list()]
 
-            is_view = True if attrs.get('tableType', 'table').lower().find('view') != -1 else False
+            table_type = attrs.get('tableType') or 'table'
+            is_view = 'view' in table_type.lower()
 
             table = Table(
                 database=table_details.get('typeName'),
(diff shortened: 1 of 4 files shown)


feng-tao:
unit test fails

metadata_service/proxy/atlas_proxy.py
```diff
@@ -450,7 +451,8 @@ def get_table(self, *, table_uri: str) -> Table:
 
             reports_guids = [report.get("guid") for report in attrs.get("reports") or list()]
 
-            is_view = True if attrs.get('tableType', 'table').lower().find('view') != -1 else False
+            table_type = attrs.get('tableType') if attrs.get('tableType') else 'table'
```
verdan, line 454: nit: @mgorsk1 you could also write this whole thing in an easy to read way

```
table_type = attrs.get('tableType') or 'table'
is_view = 'view' in table_type.lower()
```

metadata_service/proxy/atlas_proxy.py
```diff
@@ -476,6 +479,95 @@ def get_table(self, *, table_uri: str) -> Table:
...
+
+        guids = [_partition.get('guid') for _partition in _partitions
+                 if _partition.get('entityStatus') == Status.ACTIVE
+                 and _partition.get('relationshipStatus') == Status.ACTIVE]
+
+        if not guids:
+            return []
+
+        full_partitions = extract_entities(self._driver.entity_bulk(guid=list(guids), ignoreRelationships=True))
+        watermark_date_format = AtlasProxy._select_watermark_format([p.attributes.get('name') for p in full_partitions])
```
verdan, line 534: Do you really need to do this functionality here?
Since you are already validating the data on line 543, i.e., 
`partition_date, _ = AtlasProxy._validate_date(partition_name, watermark_date_format)` 
and on line 545 already making a check if partition_date, you can also make the check for watermark_date_format there if you want. 

This seems like an extra effort and extra loop as well on the full_partition

metadata_service/proxy/atlas_proxy.py
```diff
@@ -476,6 +479,95 @@ def get_table(self, *, table_uri: str) -> Table:
...
+
+                    partition_create_time = datetime.datetime.fromtimestamp(
+                        _partition_create_time).strftime(partition_value_format)
+
+                    common_values = {'partition_value': datetime.datetime.strftime(partition_date,
+                                                                                   partition_value_format),
+                                     'create_time': partition_create_time,
+                                     'partition_key': partition_key}
+
+                    if (_low_watermark_date is None) or (partition_date < _low_watermark_date):
```
verdan, line 556: Just use min() or max() to find the earliest or latest dates from a list. 
Append all the partition_dates in a list and apply min() and max() functions to get the `low_watermark` and `high_watermark`

verdan (changes requested):

metadata_service/proxy/atlas_proxy.py
mgorsk1: I think that this is actually more efficient. The assumption is that we support any number different `WATERMARK_DATE_FORMATS` but a particular table should have consistent partition naming convention.

So table A might have partitions aligning to %Y%m%d format and tables B partitions would adhere to %Y,%m format. And if table A has some partitions aligning to %Y%m%d and some to %Y-%m-%d then we should ignore the latter ones.

After making this assumption - establishing this watermark_date_format in 534 acutally means that in 543 we don't iterate through all supported date formats for each partition name, just check if the name adheres to the particular one.

verdan (approved):
Looks good to me. 💯

PR #205, opened 2020-09-28
feat: return column level badges

allisonsuarez:
### Summary of Changes
Implement functionality to add badges to columns in metadata.

### Tests

### Documentation

### CheckList

- [ ] PR passes `make test`

### DIFF
--- a/metadata_service/proxy/neo4j_proxy.py
+++ b/metadata_service/proxy/neo4j_proxy.py
@@ -120,7 +120,9 @@ class Neo4jProxy(BaseProxy):
         OPTIONAL MATCH (tbl)-[:DESCRIPTION]->(tbl_dscrpt:Description)
         OPTIONAL MATCH (col:Column)-[:DESCRIPTION]->(col_dscrpt:Description)
         OPTIONAL MATCH (col:Column)-[:STAT]->(stat:Stat)
-        RETURN db, clstr, schema, tbl, tbl_dscrpt, col, col_dscrpt, collect(distinct stat) as col_stats
+        OPTIONAL MATCH (col:Column)-[:HAS_BADGE]->(badge:Badge)-[:BADGE_FOR]->(col)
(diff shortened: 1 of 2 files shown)


metadata_service/proxy/neo4j_proxy.py
```diff
@@ -120,7 +119,9 @@ def _exec_col_query(self, table_uri: str) -> Tuple:
         OPTIONAL MATCH (tbl)-[:DESCRIPTION]->(tbl_dscrpt:Description)
         OPTIONAL MATCH (col:Column)-[:DESCRIPTION]->(col_dscrpt:Description)
         OPTIONAL MATCH (col:Column)-[:STAT]->(stat:Stat)
-        RETURN db, clstr, schema, tbl, tbl_dscrpt, col, col_dscrpt, collect(distinct stat) as col_stats
+        OPTIONAL MATCH (col:Column)-[:HAS_BADGE]->(badge:Badge)-[:BADGE_FOR]->(col)
```
feng-tao, line 122: `OPTIONAL MATCH (col:Column)-[:HAS_BADGE]->(badge:Badge)` is fine

allisonsuarez: I figured, my concern here is that adding 2 directional relationships but calling just 1 of the 2 relations can cause confusion. This was the case for @dikshathakur3119 while working on owners. I feel like it might be worth agreeing on a standard way of doing this so there is no confusion or nodes that end up with stale relationships.

feng-tao: her part is on the delete side. This one is for read side. I think we did a refactor before just use a single direction. This will also keep it consistent with other relationships.

I think the standard is to delete both directions rel in write part, but use single direction rel in read part.

allisonsuarez: sounds good

feng-tao (approved):

PR #206, opened 2020-09-28
fix: Add /delete bidirectional relations in case of owner

dikshathakur3119:
Signed-off-by: dikshathakur3119 <dikshathakur@lyft.com>

### Summary of Changes
Added Bidirectional relation to add/delete in case of user and owner relationship
When we add or delete an `OWNER` relation from UI, it only deletes one way relation whereas when we do it from extractor, it add bidirectional relations, which is causing some inconsistency in neo4j graph, and creating duplicate `OWNER_OF` relations. To make this consistent, bidirectional owner relations will be created by UI API as well.
 
### Tests
No new tests added

### Documentation

### CheckList

- [ ] PR passes `make test`

### DIFF
--- a/metadata_service/proxy/neo4j_proxy.py
+++ b/metadata_service/proxy/neo4j_proxy.py
@@ -552,8 +552,11 @@ class Neo4jProxy(BaseProxy):
         :return:
         """
         delete_query = textwrap.dedent("""
-        MATCH (n1:User{key: $user_email})-[r1:OWNER_OF]->(n2:Table {key: $tbl_key})-[r2:OWNER]->(n1) DELETE r1,r2
-        """)
+                MATCH (n1:User{key: $user_email}), (n2:Table {key: $tbl_key})
+                OPTIONAL MATCH (n1)-[r1:OWNER_OF]->(n2)
+                OPTIONAL MATCH (n2)-[r2:OWNER]->(n1) 
+                DELETE r1,r2
+                """)
 
         try:
             tx = self._driver.session().begin_transaction()
--- a/metadata_service/proxy/neo4j_proxy.py
+++ b/metadata_service/proxy/neo4j_proxy.py
@@ -546,7 +546,6 @@ class Neo4jProxy(BaseProxy):
                      owner: str) -> None:
         """
         Delete the owner / owned_by relationship.
-
         :param table_uri:
         :param owner:
         :return:
@@ -557,7 +556,6 @@ class Neo4jProxy(BaseProxy):
         OPTIONAL MATCH (n2)-[r2:OWNER]->(n1) 
         DELETE r1,r2
         """)
-        
         try:
             tx = self._driver.session().begin_transaction()
             tx.run(delete_query, {'user_email': owner,
--- a/metadata_service/proxy/neo4j_proxy.py
+++ b/metadata_service/proxy/neo4j_proxy.py
@@ -518,7 +518,7 @@ class Neo4jProxy(BaseProxy):
 
         upsert_owner_relation_query = textwrap.dedent("""
         MATCH (n1:User {key: $user_email}), (n2:Table {key: $tbl_key})
-        MERGE (n2)-[r2:OWNER]->(n1)
+        MERGE (n1)-[r1:OWNER_OF]->(n2)-[r2:OWNER]->(n1)
         RETURN n1.key, n2.key
         """)
 
@@ -552,7 +552,7 @@ class Neo4jProxy(BaseProxy):
         :return:
         """
         delete_query = textwrap.dedent("""
-        MATCH (n1:User{key: $user_email})<-[r1:OWNER]-(n2:Table {key: $tbl_key}) DELETE r1
+        MATCH (n1:User{key: $user_email})-[r1:OWNER_OF]->(n2:Table {key: $tbl_key})-[r2:OWNER]->(n1) DELETE r1,r2
         """)
 
         try:
@@ -960,7 +960,7 @@ class Neo4jProxy(BaseProxy):
         if relation_type == UserResourceRel.follow:
             relation = f'(usr{user_matcher})-[rel:FOLLOW]->(resource{resource_matcher})'
         elif relation_type == UserResourceRel.own:
-            relation = f'(usr{user_matcher})<-[rel:OWNER]-(resource{resource_matcher})'
+            relation = f'(resource{resource_matcher})-[r1:OWNER]->(usr{user_matcher})-[r2:OWNER_OF]->(resource{resource_matcher})'
         elif relation_type == UserResourceRel.read:
             relation = f'(usr{user_matcher})-[rel:READ]->(resource{resource_matcher})'
         else:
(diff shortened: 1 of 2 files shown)


feng-tao:
CI fails

metadata_service/proxy/neo4j_proxy.py
```diff
@@ -518,7 +518,7 @@ def add_owner(self, *,
 
         upsert_owner_relation_query = textwrap.dedent("""
         MATCH (n1:User {key: $user_email}), (n2:Table {key: $tbl_key})
-        MERGE (n2)-[r2:OWNER]->(n1)
+        MERGE (n1)-[r1:OWNER_OF]->(n2)-[r2:OWNER]->(n1)
```
feng-tao, line 521: i don't think you need to have bidirection in this query?

metadata_service/proxy/neo4j_proxy.py
```diff
@@ -1158,11 +1159,16 @@ def delete_resource_relation_by_user(self, *,
                                                                       user_key=user_id,
                                                                       id=id
                                                                       )
-
-        delete_query = textwrap.dedent("""
-        MATCH {rel_clause}
-        DELETE rel
-        """.format(rel_clause=rel_clause))
+        if relation_type == UserResourceRel.own:
```
feng-tao, line 1162: could you comment on why we only delete two direction for owning resources?

feng-tao: i c, now wouldn't be an issue for bookmark/follow relationships? to do it generic, let's add bi-direction for all three relationships, and do delete for r1, r2 for all three cases? WDYT?

dikshathakur3119: Yes it would be good to have consistent relations for follow and bookmark as well. I will update it

metadata_service/proxy/neo4j_proxy.py
dikshathakur3119: I think we would need bidirectional relation here as we are adding new relation between a user and table. Else we will again create duplicate relationships.
e.g From UI we created `Table-OWNER->User` relationship.  Now from backend we again create relationship between same table and user as 'Table-OWNER->User-OWNER_OF->Table', at this time neo4j graph will not find whole bidirectional relation hence it will create a pair of `OWNER and OWNER_OF ` relation, which will result into two `OWNER` relation and one `OWNER_OF` relationship between same user and table.
This is one case I could think of, but I feel there may be other cases as well because `MERGE` in neo4j graph looks for complete match.

feng-tao:
lgtm, thanks @dikshathakur3119 , have you had the chance to deploy this branch to staging for a sanity test?

feng-tao (approved):
lgtm if done a sanity test on Lyft staging

dikshathakur3119:
> lgtm if done a sanity test on Lyft staging

Deploying this branch on staging currently. Will merge after sanity tests.

feng-tao:
any other issues?

dikshathakur3119:
> any other issues?

No I did not see any other issue today. Should I merge it or wait till Monday?

metadata_service/proxy/neo4j_proxy.py
```diff
@@ -958,11 +959,14 @@ def _get_user_resource_relationship_clause(relation_type: UserResourceRel, id: s
                 user_matcher += ' {key: $user_key}'
 
         if relation_type == UserResourceRel.follow:
-            relation = f'(usr{user_matcher})-[rel:FOLLOW]->(resource{resource_matcher})'
+            relation = f'(resource{resource_matcher})-[r1:FOLLOWED_BY]->(usr{user_matcher})-[r2:FOLLOW]->' \
```
PaschalisDim, line 962: I think those lines should include optional as in the case of deletion. If not then, the existing owners/bookmarks, will not be able to retrieve their existing ones, if they have just one relation and thus it is a breaking change.

feng-tao: yeah, we should make it option and delete if so. cc @dikshathakur3119 wdyt?

dikshathakur3119: Thanks @PaschalisDim for mentioning, this function is used to create query for add, delete and read, I will fix this function to return appropriate match for each of these cases.

PaschalisDim: Thanks for taking this over.
@dikshathakur3119 
Creation and deletion I guess should be fine. However, when returning what exists, we need to take care to return the ones that were created prior to this MR, so that is the only thing to get altered to include all cases.
Thanks

dikshathakur3119: No actually, that was reason I made this change at first place. If we do delete with existing query, that is  `A -Relation1->B-Realtion2->A` , it only deletes if there is exact match for bidirectional relation, else it will not make any delete to single relations. And when you use single directional queries to read a relation it will show your deleted tables as well.
Similarly, in case of add, if we add with query `A -Relation1->B-Realtion2->A`, it will create duplicate relation if there was existing only one way relation. And this will ultimately cause a loop and you will never be able to delete any owner/bookmark which was added from Frontend UI before this MR .

dikshathakur3119: @PaschalisDim let me know if it makes sense

PaschalisDim: Hi @dikshathakur3119 
Apologies, I was not clear enough here.
So, before this MR let's assume that there is a user who owns and follows a table (table A).
So he will have one relation (usr{user_matcher})<-[rel:OWNER]-(resource{resource_matcher}) and one relation
(usr{user_matcher})-[rel:FOLLOW]->(resource{resource_matcher}).
Then you make this MR.
After we merge this MR and it goes to production,
when the user will try the tables that he owns, the amundsenmeta library will search for (resource{resource_matcher})-[r1:OWNER]->(usr{user_matcher})-[r2:OWNER_OF]->' \
                       f'(resource{resource_matcher}) 
these type of matches in order to see whether the user owns any table.
Since before this MR the user just had one-directional, the tableA will not show up in the list of the owned tables.
If he adds himself as an owner to another table, tableB, then he will have applied right the bi-directional and he will just see tableB in the list of owned tables.
To the point, for the existing one-way relations this change is a breaking change.
The add functionality should be fine. The get functionality is in my thoughts not right, since it will get just owners that have assigned themselves as owners just only AFTER this MR was deployed while I am not sure about the delete one which needs to be tested after the GET functionality is fixed.
@dikshathakur3119

PaschalisDim:
@dikshathakur3119 
Another thing that came to my mind today is that we need to make sure that we also update the neo4j_staleness_removal_task from databuilder when this one gets finalized, to ensure deletion of all the possible relationships of the table, since after this MR there will be more relationships added.
Thanks!

tests/unit/proxy/test_neo4j_proxy.py
```diff
@@ -963,28 +963,32 @@ def test_user_resource_relation_clause(self) -> None:
...
+            expected = '(resource:Table {key: $resource_key})-[r1:READ_BY]->(usr:User {key: $user_key})-[r2:READ]->' \
+                       '(resource:Table {key: $resource_key})'
             self.assertEqual(expected, actual)
 
             actual = neo4j_proxy._get_user_resource_relationship_clause(UserResourceRel.own,
                                                                         id='foo',
                                                                         user_key='bar',
                                                                         resource_type=ResourceType.Table)
-            expected = '(usr:User {key: $user_key})<-[rel:OWNER]-(resource:Table {key: $resource_key})'
+            expected = '(resource:Table {key: $resource_key})-[r1:OWNER]->(usr:User {key: $user_key})-[r2:OWNER_OF]->' \
```
pPanda-beta, line 982: @verdan 
We need something  similar (and precise as well) for atlas as well, what do you think?

PR #216, opened 2020-10-15
fix: update bookmark qualified name regex related to #740

pPanda-beta:
TODO: Need to add unit tests

### Summary of Changes

User ids can hold anything, like special characters [ '.', '@' ]. So changed the userid regex.

### Tests

Could not find existing test cases, need to setup a big fixture to unit test this.
Some quick test cases could be : 
image

### CheckList

- [x] PR passes `make test`

### DIFF
--- a/metadata_service/proxy/atlas_proxy.py
+++ b/metadata_service/proxy/atlas_proxy.py
@@ -169,7 +169,7 @@ class AtlasProxy(BaseProxy):
         \.
         (?P<entity_type>[^.]*)
         \.
-        (?P<user_id>[^.]*)\.bookmark
+        (?P<user_id>.*)\.bookmark
         \@
         (?P<cluster>.*)
         $


feng-tao:
cc @verdan @mgorsk1

mgorsk1:
doesn't current regex enable just that ? it's [^.]* so basically anything ? @pPanda-beta

pPanda-beta:
@mgorsk1 
I tried on https://regex101.com/ and as well as python3.7 shell. It doesn’t work in that way.
I believe  [^.]* means 
ZERO OR MORE (*)
ANY CHARACTER 
EXCEPT (^)
DOT (.)

pPanda-beta:
#keep_fresh

pPanda-beta:
@mgorsk1 any updates on this?

pPanda-beta:
@mgorsk1 Since we are using greedy extraction any dots in between will be consumed as part of the username only. Check my 3rd example. "lover.bookmark@hactoberfest.do" is captured as username. Nothing more nothing less. 

### Why am I using dots in username?
In Rapido's dataplatform we use keycloak for authentication. When using any identity provider with keycloak (like google, fb, github,...) it mostly takes email as the username. Our authorization is also powered by apache ranger where every user's username is their email id. This works for both presto and apache atlas.

verdan:
I think we should allow dots in user_id. Either via moving away from the naming conventions approach or simply using the relationships. The relationships won't be backward compatible but more efficient. In any case, this should be fixed.

pPanda-beta:
@verdan I guess we can still go with the current naming conventions approach. 

`db.table_name.entity_type.USER_ID_WITH_DOTS_AND_@_SYMBOL.bookmark@primary`

The good point is,  any component apart from user id does not contain dots or '@' symbol. So playing cool regex games we can still continue on this.


On the relationship approach: 
1. It is good, and simpler.
2. We can have a window of 6 months with both the approaches, to allow users to migrate.

mgorsk1:
I am all-in for relying on relationships, not naming conventions. We've already experienced issues with naming convention approach on frequent users feature which was main motivation to extend it with reader <> table relationship and I think the same approach would work well with bookmarks. 

Most of the code is already there (very similar thing is done in data owner methods), we just need to extend amundsenatlastypes. 

Only pain-point would be adding relationships to bookmarks already present in Amundsen & Atlas deployment.

How many bookmarks you have in your setup @pPanda-beta ? 6 months seems like a long time for such minor adjustment.

pPanda-beta:
@mgorsk1 
No worries about my setup, before installing new amundsenatlas types, we will create scripts to backup, modify and restore existing entities. 

I suggested 6 months considering other consumer base of amundsen.
For us, everything is protected by tests and undergo through lower environments for manual testing. So we can safely do it overnight.

verdan:
@pPanda-beta can you please try to fix this up using the relationship approach?

pPanda-beta:
@verdan Yes that can be done but not related with this PR.
This solves the problem of current approach. 

To solve it using relationships I suggest two alternative methods. 

1. Two API calls. 
  i. Call atlas to find the bookmarks of current logged in user. - Basic Search
  ii. Call atlas to find the entities of those bookmarks - Either basic search or "bulk entity by guid"
 
  Pros: Fast, 
  Cons: Multiple places of failure, too much code for simple task. Programmatically doing basic search is too much of code.
  
2. Single API call
   Use the dsl search 
   query : `from Bookmark where user.qualifiedName='palash.das@rapido.bike'  select entity` (tested OK)
   
   Pros: Pretty straight forward, simple to implement
   Cons: Slow, every dsl search spins up a new gremlin script engine
   
We need to brainstorm more to decide one of the approaches. Till that time I was requesting this bug fix, since it solves the problem of current qualified name based approach.

dorianj:
Please re-open either as an RFC to clear up the implementation options, or a new PR against the monorepo when it's ready.

PR #232, opened 2020-11-13
feat: Updated popular_tables endpoint to allow optional user_id

joshthoward:
Signed-off-by: Josh Howard <joshthoward@gmail.com>

### Summary of Changes

The popular_tables endpoint has been updated to allow for an optional user_id. The idea is that the popular tables response should be able to be user specific. 

### Tests

Minor updates to fix `assert_called_with` calls. 

### CheckList

- [-] PR title addresses the issue accurately and concisely. Example: "Updates the version of Flask to v1.0.2"
- [-] PR includes a summary of changes.
- [-] PR adds unit tests, updates existing unit tests, **OR** documents why no test additions or modifications are needed.
- [-] In case of new functionality, my PR adds documentation that describes how to use it.
- [-] PR passes `make test`

### DIFF
--- a/metadata_service/__init__.py
+++ b/metadata_service/__init__.py
@@ -88,7 +88,9 @@ def create_app(*, config_module_class: str) -> Flask:
 
     api = Api(api_bp)
 
-    api.add_resource(PopularTablesAPI, '/popular_tables/')
+    api.add_resource(PopularTablesAPI,
+                     '/popular_tables/',
+                     '/popular_tables/<path:user_id>')
     api.add_resource(TableDetailAPI, '/table/<path:table_uri>')
     api.add_resource(TableDescriptionAPI,
                      '/table/<path:id>/description')
--- a/metadata_service/api/popular_tables.py
+++ b/metadata_service/api/popular_tables.py
@@ -2,7 +2,7 @@
 # SPDX-License-Identifier: Apache-2.0
 
 from http import HTTPStatus
-from typing import Iterable, List, Mapping, Union
+from typing import Iterable, List, Mapping, Union, Optional
 
 from amundsen_common.models.popular_table import (PopularTable,
                                                   PopularTableSchema)
@@ -21,8 +21,9 @@ class PopularTablesAPI(Resource):
         self.client = get_proxy_client()
 
     @swag_from('swagger_doc/popular_tables_get.yml')
-    def get(self) -> Iterable[Union[Mapping, int, None]]:
+    def get(self, user_id: Optional[str] = None) -> Iterable[Union[Mapping, int, None]]:
         limit = request.args.get('limit', 10, type=int)
-        popular_tables: List[PopularTable] = self.client.get_popular_tables(num_entries=limit)
+        popular_tables: List[PopularTable] = self.client.get_popular_tables(num_entries=limit,
+                                                                            user_id=user_id)
         popular_tables_json: str = PopularTableSchema(many=True).dump(popular_tables).data
         return {'popular_tables': popular_tables_json}, HTTPStatus.OK
--- a/metadata_service/proxy/atlas_proxy.py
+++ b/metadata_service/proxy/atlas_proxy.py
@@ -755,7 +755,9 @@ class AtlasProxy(BaseProxy):
 
         return popular_tables
 
-    def get_popular_tables(self, *, num_entries: int) -> List[PopularTable]:
+    def get_popular_tables(self, *,
+                           num_entries: int,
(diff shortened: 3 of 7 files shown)


markgrover:
Hi @joshthoward, thanks for this. Appreciate your contribution to Amundsen and apologies for the delay in responding.

I think this is a step in the right direction. Often what's popular in the whole company doesn't matter to an analyst in one team. What's popular in their team, or what's popular personalized to them is more relevant.

Perhaps, I missed it but this code seems to add the end point to allow the `user_id` but I don't have a good sense of how that user_id will be used. Can you help me understand how you are thinking about that, or if you already have a working example of that?

joshthoward:
@markgrover You didn't miss anything in the code. I didn't implement a backend query using user ID in this PR for two reasons... 
1) I wanted the PR to be considered on the basis of "Should the popular tables response be able to be personalized by a backend" rather than my specific heuristic choice
2) From my other [PR](https://github.com/amundsen-io/amundsenfrontendlibrary/pull/740), it seems like the jury is still out on how customizations should be made (shameless plug... I would advocate to do everything through environment variables with hooks in the helm charts, but we'll see).

To answer your question on how to use it, a user specific heuristic for Neo4j could be to "Return the top `k` tables viewed by users who viewed the same tables as the current user". Pseudo cypher code is below:

```
MATCH (viewer: User {key: '<user_id>'})-[:viewed]->(table1: Table)<-[:viewed]-(coViewer: User)-[:viewed]->(table2: Table)
WHERE table1 <> table2
RETURN table2.name, count(table2) as frequency
ORDER BY frequency DESC
LIMIT <limit> 
```

The implementation would look like the following in `metadata_service/proxy/neo4j_proxy.py`

```python
if user_id is not None: 
    # run query above
else: 
    # run old query
```

Please let me know your thoughts!

markgrover:
I love this, this is super helpful @joshthoward. I could see personalization happen on a per-team basis as well e.g. show the most queried tables used by this team as a whole. But I don't see why that needs to block this.

So, I am supportive of this. In terms of how customizations should be made, I don't have an opinion on, right now. I noticed that @feng-tao already requested some maintainers to look, I will ping them again and will take a look myself too.


This is a +1 from my side, let me know if anyone else has any feedback on this. cc'ing some people who may: @feng-tao @Golodhros @verdan

joshthoward:
@markgrover 

Well thanks!

Based on your comment "I could see personalization happen on a per-team basis as well e.g. show the most queried tables used by this team as a whole.", I am going to close the [PR](https://github.com/amundsen-io/amundsenfrontendlibrary/pull/765) on the frontend until we best figure out how to display both.

markgrover:
Sounds good. I see popularity calculation evolving and that's ok. Thinking
more, I like the per person approach better since there may be other people
not in your team, plus team data is pretty messy in many organizations and
there's no standard.

So, all in all, I am all in favor of being scrappy and moving forward here.

feng-tao:
let me take a look tonight.

feng-tao:
@joshthoward , lgtm in high level. A fe comments:
1. given we take the user_id as input, i think we should just change the popular table query (neo4j) to take the user id into account if it is provided and remain the same if it is None.
2. For FE, I think we should set it through a config to allow user to change it whether to display global popular table vs personal popular table. The reason is that for new hires, they may not have frequent used tables which they could use the global popular table for reference.

wdyt?

joshthoward:
@feng-tao It would be my recommendation to move forward with this PR as is. 

1) I agree, but I want to do further experimentation on building a real recommender system. I don't have users to experiment with, but if someone could run the following query on a production Neo instance, that would give me what I need for that: 
`MATCH (n:User)-[FOLLOW]->(t:Table) RETURN apoc.util.sha256([n.key]),apoc.util.sha256([t.key])`
2) I'm thinking that the config should go on the frontend because a user might want global views, personalized views, or both which could result in a frontend change.

feng-tao:
> @feng-tao It would be my recommendation to move forward with this PR as is.
> 
> 1. I agree, but I want to do further experimentation on building a real recommender system. I don't have users to experiment with, but if someone could run the following query on a production Neo instance, that would give me what I need for that:
>    `MATCH (n:User)-[FOLLOW]->(t:Table) RETURN apoc.util.sha256([n.key]),apoc.util.sha256([t.key])`
> 2. I'm thinking that the config should go on the frontend because a user might want global views, personalized views, or both which could result in a frontend change.

yeah, when i said the config, I mean FE config (the PR you closed).

feng-tao (approved):

PR #248, opened 2021-01-27
feat: Enable custom keyword arguments in Atlas Proxy Client

mgorsk1:
### Summary of Changes

This PR adds a possibility to pass custom kwargs to Atlas proxy client. It be used to fine-tune parameters like timeout or num of retries. Since every client might accept different kind of extra parameters it might be easier to just have custom dictionary complementing 

### CheckList

- [x] PR passes `make test`

### DIFF
--- a/metadata_service/config.py
+++ b/metadata_service/config.py
@@ -17,6 +17,7 @@ PROXY_PASSWORD = 'PROXY_PASSWORD'
 PROXY_ENCRYPTED = 'PROXY_ENCRYPTED'
 PROXY_VALIDATE_SSL = 'PROXY_VALIDATE_SSL'
 PROXY_CLIENT = 'PROXY_CLIENT'
+PROXY_CLIENT_KWARGS = 'PROXY_CLIENT_KWARGS'
 
 PROXY_CLIENTS = {
     'NEO4J': 'metadata_service.proxy.neo4j_proxy.Neo4jProxy',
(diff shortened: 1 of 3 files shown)


mgorsk1:
cc @dechoma

dechoma (commented):
lgtm

metadata_service/config.py
```diff
@@ -84,6 +85,10 @@ class Config:
     # format to be used in tables
     WATERMARK_DATE_FORMATS = ['%Y%m%d']
 
+    # Custom kwargs that will be passed to proxy client. Can be used to fine-tune parameters like timeout
+    # or num of retries
+    PROXY_CLIENT_KWARGS: Dict = dict()
```
feng-tao, line 90: do you have example of this config? and is it possible to be leverage by other proxies?

mgorsk1: It definitely can be leveraged by different proxies. Two examples:

- In neo4j proxy there are settings that clearly are not set by this singleton (https://github.com/amundsen-io/amundsenmetadatalibrary/blob/master/metadata_service/proxy/neo4j_proxy.py#L54). I think that currently users don't have any way of fine tuning those. but they could be easily passed under this new configuration option.
- Some time ago we made a POC with GCP Data Catalog as Amundsen proxy and GCP client has completely different set of args in comparison to what Neo4j, Atlas or Neptune expects. We used similar solution back then.

In our case we started small with `PROXY_CLIENT_KWARGS = dict(timeout=int(os.getenv('ATLAS_REQUEST_TIMEOUT_SEC', 30)))` but there are couple of more params we could

mgorsk1:
@feng-tao ok to accept & merge?

feng-tao (approved):

mgorsk1:
Thanks ! @feng-tao

metadata_service/proxy/atlas_proxy.py
```diff
@@ -70,7 +70,8 @@ def __init__(self, *,
                  user: str = 'admin',
                  password: str = '',
                  encrypted: bool = False,
-                 validate_ssl: bool = False) -> None:
+                 validate_ssl: bool = False,
+                 client_kwargs: dict = dict()) -> None:
```
feng-tao, line 74: should use None as default argument.

feng-tao:
@mgorsk1 could you take a look at this issue (https://github.com/amundsen-io/amundsen/issues/881) ? Looking at the code, if people doesn't define the config, I think it will break other proxy (neo4j, neptune etc)?

Issue #13, opened 2019-03-19
Documentation on how to make Apache Atlas compatible with Amundsen


verdan:
I'll be writing a complete documentation on how to setup the Apache Atlas in order to make it compatible with the Apache Atlas proxy in the amundsenmetadata library. 

We can actually handle that pretty easily and efficiently by adding a superType, something like `AmundsenTable` etc, and assign that to only those tables/data sources that we need in the amundsen. (hence not showing the DBs in amundsen but just the tables)

verdan:
This already has been implemented and merged in master branch:
https://github.com/lyft/amundsenmetadatalibrary/blob/master/docs/proxy/atlas_proxy.md

PR #270, opened 2021-03-03
fix: configs and imports for Gremlin and Neptune Config

verdan:
Signed-off-by: verdan <verdan.mahmood@gmail.com>

### Summary of Changes

Fixes: https://github.com/amundsen-io/amundsen/issues/937

### Tests

### Documentation

### CheckList

- [ ] PR passes `make test`

### DIFF
--- a/metadata_service/config.py
+++ b/metadata_service/config.py
@@ -104,13 +104,6 @@ class LocalConfig(Config):
 
 
 class AtlasConfig(LocalConfig):
-    # Used to differentiate tables with other entities in Atlas. For more details:
-    # https://github.com/amundsen-io/amundsenmetadatalibrary/blob/master/docs/proxy/atlas_proxy.md
-    ATLAS_TABLE_ENTITY = 'Table'
-
-    # The relationalAttribute name of Atlas Entity that identifies the database entity.
-    ATLAS_DB_ATTRIBUTE = 'db'
-
     # List of accepted date formats for AtlasProxy Watermarks. With this we allow more than one datetime partition
     # format to be used in tables
     WATERMARK_DATE_FORMATS = ['%Y%m%d']
--- a/metadata_service/config.py
+++ b/metadata_service/config.py
@@ -53,13 +53,6 @@ class Config:
 
     IS_STATSD_ON = False
 
-    # Used to differentiate tables with other entities in Atlas. For more details:
-    # https://github.com/amundsen-io/amundsenmetadatalibrary/blob/master/docs/proxy/atlas_proxy.md
-    ATLAS_TABLE_ENTITY = 'Table'
-
-    # The relationalAttribute name of Atlas Entity that identifies the database entity.
-    ATLAS_DB_ATTRIBUTE = 'db'
-
     # Configurable dictionary to influence format of column statistics displayed in UI
     STATISTICS_FORMAT_SPEC: Dict[str, Dict] = {}
 
@@ -80,10 +73,6 @@ class Config:
 
     # List of regexes which will exclude certain parameters from appearing as Programmatic Descriptions
     PROGRAMMATIC_DESCRIPTIONS_EXCLUDE_FILTERS = []  # type: list
-
-    # List of accepted date formats for AtlasProxy Watermarks. With this we allow more than one datetime partition
-    # format to be used in tables
-    WATERMARK_DATE_FORMATS = ['%Y%m%d']
 
     # Custom kwargs that will be passed to proxy client. Can be used to fine-tune parameters like timeout
     # or num of retries
@@ -112,6 +101,19 @@ class LocalConfig(Config):
         'title': 'Metadata Service',
         'uiversion': 3
     }
+
+
+class AtlasConfig(LocalConfig):
+    # Used to differentiate tables with other entities in Atlas. For more details:
+    # https://github.com/amundsen-io/amundsenmetadatalibrary/blob/master/docs/proxy/atlas_proxy.md
+    ATLAS_TABLE_ENTITY = 'Table'
+
+    # The relationalAttribute name of Atlas Entity that identifies the database entity.
+    ATLAS_DB_ATTRIBUTE = 'db'
+
+    # List of accepted date formats for AtlasProxy Watermarks. With this we allow more than one datetime partition
+    # format to be used in tables
+    WATERMARK_DATE_FORMATS = ['%Y%m%d']
 
 
 class GremlinConfig(LocalConfig, LocalGremlinConfig):
--- a/metadata_service/config.py
+++ b/metadata_service/config.py
@@ -119,10 +119,10 @@ class GremlinConfig(LocalConfig, LocalGremlinConfig):
 
 
 class NeptuneConfig(LocalGremlinConfig, LocalConfig):
-    import boto3
+    import boto3    # isort:skip
 
     def __init__(self) -> None:
-        from amundsen_gremlin.test_and_development_shard import shard_set_explicitly
+        from amundsen_gremlin.test_and_development_shard import shard_set_explicitly    # isort:skip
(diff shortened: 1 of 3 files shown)


requirements.txt
```diff
@@ -47,7 +47,6 @@ gremlinpython==3.4.3
 aniso8601==3.0.0
 attrs==19.1.0
 click==6.7
-gremlinpython==3.4.3
```
verdan, line 50: duplicate of line # 46.

metadata_service/config.py
```diff
@@ -118,37 +114,30 @@ class LocalConfig(LocalGremlinConfig, Config):
...
-class NeptuneConfig(LocalGremlinConfig, Config):
     DEBUG = False
-    TESTING = False
     LOG_LEVEL = 'INFO'
-    LOCAL_HOST = '0.0.0.0'
 
     # FORMAT: wss://<NEPTUNE_URL>:<NEPTUNE_PORT>/gremlin
     PROXY_HOST = os.environ.get('PROXY_HOST')
     PROXY_PORT = None
     PROXY_CLIENT = PROXY_CLIENTS['NEPTUNE']
```
verdan, line 136: This class is inherited from Config, and these values should be set via the environment variable available instead of redeclaring them for different proxies. It will be easy to manage different proxies this way if we change any variable/config

metadata_service/config.py
```diff
@@ -93,7 +91,7 @@ class Config:
 
 
 # NB: If you're using the gremlin proxy, the appropriate GremlinConfig must be added to any other configs
-class LocalConfig(LocalGremlinConfig, Config):
+class LocalConfig(Config):
```
verdan, line 94: Should not be using `LocalGremlinConfig` if not required.

feng-tao:
lgtm, but will let @AndrewCiambrone takes another look. Also could we do the same for atlas proxy ? Currently there are a few configs in the base class which are not related to other proxy types.

metadata_service/config.py
```diff
@@ -118,37 +114,30 @@ class LocalConfig(LocalGremlinConfig, Config):
...
+    JANUS_GRAPH_URL = None
+
+
+class NeptuneConfig(LocalGremlinConfig, LocalConfig):
+    import boto3    # isort:skip
 
+    def __init__(self) -> None:
+        from amundsen_gremlin.test_and_development_shard import shard_set_explicitly    # isort:skip
+        # The databuilder expects this to be False currently. We are defaulting to true because the testing expects this
+        if bool(distutils.util.strtobool(os.environ.get('IGNORE_NEPTUNE_SHARD', 'False'))):
```
AndrewCiambrone, line 127: Github might be showing this weird but this will not work inside the init. 

The set_shard will only work when it has not been used yet. 
https://github.com/amundsen-io/amundsengremlin/blob/master/amundsen_gremlin/test_and_development_shard.py#L53

and as pointed out by the issue 
https://github.com/amundsen-io/amundsengremlin/blob/master/amundsen_gremlin/test_and_development_shard.py#L74

makes it used.

AndrewCiambrone:
Overall I think this looks great. I am a big fan of separating the configs. The only concern I have is placing the `shard_set_explicitly` inside the initializing function of the config.

metadata_service/config.py
mgorsk1: +1 for this approach, we should never rely on functions to be called implicitly on imports anyways

verdan: @AndrewCiambrone I think this will work as expected. Would be nice if you can please give this a try. 
On the other hand, I believe you already answered yourself how to fix this properly. these kinds of functions should not be called globally and may cause side effects like these.

metadata_service/config.py
```diff
@@ -118,37 +103,43 @@ class LocalConfig(LocalGremlinConfig, Config):
     }
 
 
-# The databuilder expects this to be False currently. We are defaulting to true because the testing expects this
-if bool(distutils.util.strtobool(os.environ.get('IGNORE_NEPTUNE_SHARD', 'False'))):
-    shard_set_explicitly('')
+class AtlasConfig(LocalConfig):
```
mgorsk1, line 106: let me know wdyt @verdan

verdan: sweeet! 💯

mgorsk1: removed everything except `WATERMARK_DATE_FORMATS` as I think table and db entity names should not be configurable anyways. so it makes Atlas Proxy follow regular configuration in 99%

verdan: @mgorsk1 let's try to make `WATERMARK_DATE_FORMATS` generic as well for all proxies, and not make it specific to Atlas. This will remove the AtlasConfig altogether.

mgorsk1: I am not sure if design of databuilder/watermak feature on neo4j side actually requires this, it's very Atlas specific

metadata_service/config.py
AndrewCiambrone: Hey @verdan I pulled down the branch and tried it locally. It unfortunately did not work. It looks like the `__init__ `is not called upon. 
Digging in deeper it looks like because the config is never initialized inside the from_object function: https://github.com/pallets/flask/blob/master/src/flask/config.py#L161

AndrewCiambrone: I am not familiar enough with the test_and_development_shard.py to know if we can just remove `_reset_for_testing_only()` from being called implicitly. Maybe @friendtocephalopods  might know? 

But a short term solution could be:
- To set the default of `IGNORE_NEPTUNE_SHARD` to be `True` 
- Or move the Neptune config to its own file.

verdan: @mgorsk1 @AndrewCiambrone how do you think we should proceed with this? 
For a quick fix, we can do this in following ways:
- First, as Andrew mentioned, to set this to True. 
- Second, `if os.environ.get('PROXY_CLIENT', 'NEO4J') == 'NEPTUNE': then the check, and import shard_set_explicitly('')`

Thoughts?

mgorsk1: My thoughts are we should remove this line https://github.com/amundsen-io/amundsengremlin/blob/master/amundsen_gremlin/test_and_development_shard.py#L74 , add explicit call in tests (where it's actually requried) and bump gremlin dependency. But I don't know this proxy well enough to asses what would be the impact of removing this line.

AndrewCiambrone: @mgorsk1 @verdan Until someone someone who is more familiar with the testing framework can vouch that removing line 74 is okay I am proposing: https://github.com/amundsen-io/amundsengremlin/pull/13

friendtocephalopods: Sure, that change (#13) should be fine! You don't _need_ to shard anything here by user; it's just convenient if you want to have one neptune instance backing multiple tests at once. Defaulting to some static value works.

metadata_service/config.py
feng-tao: we don't need `WATERMARK_DATE_FORMATS` in the databuilder AFAIK.

feng-tao:
@verdan @mgorsk1 what is the next step for the pr?

verdan:
I am not in favor of calling `shard_set_explicitly` on the global config level within the Amundsen config. 
@AndrewCiambrone do you think you can call or move this test/local only config somewhere in the amundsengremlin repo?

AndrewCiambrone:
I was really hoping for a config that worked without too many code changes. So that a a user could get the metadata service up and running by only setting a few environment variables. (just like they can with Atlas and Neo4j). I am fairly confident the original issue is solved if we bump the amundsengremlin version. I do think we should create another issue  to clean up the test_and_development_shard.py file. Which would allow us to remove the global `shard_set_explicitly`.

AndrewCiambrone:
@verdan I think I have a solution that works for everyone. Waiting to see if the approach is okay from @friendtocephalopods.

PR #273, opened 2021-03-06
feat: Add Column Badge API

alldoami:
### Summary of Changes

Added Column Badge API that addresses this issue: amundsen-io/amundsen#944

### Tests

Tested exactly how we're testing Table Badges right now.

### Documentation

https://github.com/amundsen-io/amundsen/pull/962

### CheckList

- [ ] PR passes `make test`


metadata_service/proxy/neo4j_proxy.py
```diff
@@ -592,14 +592,20 @@ def add_badge(self, *,
...
 
         LOGGER.info('New badge {} for id {} with category {} '
                     'and resource type {}'.format(badge_name, id, category, resource_type.name))
 
         validation_query = \
             'MATCH (n:{resource_type} {{key: $key}}) return n'.format(resource_type=resource_type.name)
 
+        column_suffix = ''
+        if column_name:
+            column_suffix = '/' + column_name
```
allisonsuarez, line 606: why not just reassign id here rather than add the + '' in case where column doesn't exist? Also can use interpolation rather than '/' + var so it's consistent with other code

alldoami: Ya I didn't know if it was best to change the input variable as others might want to use the variable for something else.

alldoami:
Would love to get Dmitriy's code to get merged fixing https://github.com/amundsen-io/amundsen/issues/939 before merging this.

dkunitsk:
amundsen-io/amundsen#939 should be fixed now (by https://github.com/amundsen-io/amundsen/pull/951)

alldoami:
Thanks @dkunitsk! Would love to get this in the new release then 😄

metadata_service/proxy/neo4j_proxy.py
```diff
@@ -655,6 +661,10 @@ def delete_badge(self, id: str,
         [r1:BADGE_FOR]->(n:{resource_type} {{key: $key}})-[r2:HAS_BADGE]->(b) DELETE r1,r2
         """.format(resource_type=resource_type.name))
 
+        if column_name:
+            id = '{}/{}'.format(id, column_name)
+            LOGGER.info('ID with colum name is now {}'.format(id))
```
dorianj, line 666: remove extraneous log lines (same above)

metadata_service/proxy/neo4j_proxy.py
```diff
@@ -592,14 +592,19 @@ def add_badge(self, *,
                   id: str,
                   badge_name: str,
                   category: str = '',
-                  resource_type: ResourceType = ResourceType.Table) -> None:
+                  resource_type: ResourceType,
+                  column_name: str = '') -> None:
```
dorianj, line 596: I think it would be better to have the callers be responsible for generating the correctly-formed id, rather than passing in `column_name`? That would generalize to badging other types of content better.

If adopted, would be sensible to have a helper method that generates the id from a column name and id.

An alternative would be to pass in a new type of object that encapsulates `ResourceType` and `id`, and is responsible for containing the state and logic needed to generate id.

wdyt?

dorianj (changes requested):
Thanks for the contribution!

metadata_service/api/column.py
```diff
@@ -60,3 +62,35 @@ def get(self, table_uri: str, column_name: str) -> Union[tuple, int, None]:
 
         except Exception:
             return {'message': 'Internal server error!'}, HTTPStatus.INTERNAL_SERVER_ERROR
+
+
+class ColumnBadgeAPI(Resource):
```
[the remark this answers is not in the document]
dorianj: it looks like it's mostly dashboard code that's using that capability -- so while it's certainly option to implement it, it's encouraged since safer and better error messages

metadata_service/proxy/neo4j_proxy.py
alldoami: Through the API, the user would just do: `curl -X PUT https://{your metadata url}/table/{table key}/column/{column name}/badge/{badge name}?category={badge category}` they'd still need to include the column name somehow and we'd need to check that they want to badge a column vs a table. A `correctly-formed id` would still have to include the column name, and I think this is the most straightforward way to provide that. I can for sure add a helper function if we want to make it more clear that the `id` is changing though.

alldoami: For implementation clarification, do the API call formats need to be different for table and column badge? How does the API know when to use the different API classes? It looks like `ColumnDescriptionAPI` is also analogous to `TableDescriptionAPI` and takes in a `column_name`.

alldoami: `we're getting that info from the resource_type, right?` This is because the API call format is unique from the table badge and so we're able to figure out which class to call because of it.

metadata_service/api/column.py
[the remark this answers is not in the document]
alldoami: ignoring for now ~

metadata_service/api/swagger_doc/column/badge_put.yml
```diff
@@ -0,0 +1,47 @@
+Add badge to a resource
+---
+tags:
+  - 'table'
+  - 'dashboard'
```
dorianj, line 5: i don't know what `tags` mean here, but does this apply?

dorianj (approved):
Nice!

metadata_service/api/swagger_doc/column/badge_put.yml
alldoami: I used the same yml as the table ones, so i just kept them here... not sure what they mean...

metadata_service/api/column.py
```diff
@@ -89,3 +90,33 @@ def get(self, table_uri: str, column_name: str) -> Union[tuple, int, None]:
...
+    def __init__(self) -> None:
+        self.client = get_proxy_client()
+        self.parser = reqparse.RequestParser()
+        self.parser.add_argument('category', type=str, required=True)
+        super(ColumnBadgeAPI, self).__init__()
+
+        self._badge_common = BadgeCommon(client=self.client)
+
+    @swag_from('swagger_doc/column/badge_put.yml')
```
feng-tao, line 104: the swagger doesn't look correct. any reason we have another badge put instead of reusing `swagger_doc/badge/badge_put.yml`

feng-tao (changes requested):

metadata_service/api/column.py
```diff
@@ -89,3 +90,33 @@ def get(self, table_uri: str, column_name: str) -> Union[tuple, int, None]:
...
+    def put(self, table_uri: str, badge: str, column_name: str) -> Iterable[Union[Mapping, int, None]]:
+        args = self.parser.parse_args()
+        category = args.get('category', '')
+
+        return self._badge_common.put(id=f"{table_uri}/{column_name}",
+                                      resource_type=ResourceType.Column,
+                                      badge_name=badge,
+                                      category=category)
+
+    @swag_from('swagger_doc/column/badge_delete.yml')
```
feng-tao, line 114: same

metadata_service/api/column.py
feng-tao: cc @allisonsuarez  I think the column badge should extend the same swagger file shared by table/dashboard badge (https://github.com/amundsen-io/amundsenmetadatalibrary/blob/master/metadata_service/api/swagger_doc/badge/badge_put.yml)

alldoami: The `column` badges require another parameter (`column_name`) and it uses `table_uri` and not `id`.

feng-tao: hey @alldoami if you take a look at https://github.com/amundsen-io/amundsenmetadatalibrary/blob/master/metadata_service/api/dashboard.py#L92 or https://github.com/amundsen-io/amundsenmetadatalibrary/blob/master/metadata_service/api/table.py#L199 which both takes an id as input param which the id could be a table_uri, dashboard_uri or column_uri which could unique identifies the entity resource.  With that, we won't need table_uri, column_name.

I think we should do the same here.

alldoami: @dorianj had stronger feelings about this 😅

alldoami: We also talked about this through slack, we'll need the `column_name` as it'd be more straightforward to have the API resource be `table/<table_uri>/column/<column_name>/badge/<badge_name>/<category>` instead of `table/<table_id with column/column/badge...`. We need a unique identifier to tell us which class to use (which is why we include the `column` in the API call. If we went with `id`, it would look exactly like the table badge API and we'd need to include something like the word `column` in there. I think my way is the most straightforward still.

alldoami: We also talked about this through slack, we'll need the `column_name` as it'd be more straightforward to have the API resource be `table/<table_uri>/column/<column_name>/badge/<badge_name>/<category>` instead of `table/<table_id with column/column/badge...`. We need a unique identifier to tell us which class to use (which is why we include the `column` in the API call. If we went with `id`, it would look exactly like the table badge API and we'd need to include something like the word `column` in there. I think my way is the most straightforward still.

feng-tao: any reason the API can't be `/column/<path:id>/badge/<badge>`

alldoami: Also if we change it to using `swagger_doc/badge/badge_put.yml`, then the `id`s will be accepting different formatted values for table and column, which could also cause confusion. It's an easy fix it you want me to change it, but want to explain why we implemented this way before changing.

feng-tao: I see, thanks for sharing the context, I am ok with your format to make it consistent with the other column API.

metadata_service/api/swagger_doc/column/badge_delete.yml
```diff
@@ -0,0 +1,47 @@
+Delete badges of a resource
+---
+tags:
+  - 'table'
+  - 'dashboard'
```
feng-tao, line 5: but why the tag shows table, dashboard? should it just show column?

feng-tao (approved):

feng-tao:
thanks @alldoami

alldoami:
Thanks @feng-tao! Should we merge https://github.com/amundsen-io/amundsen/pull/962 too? I fixed the merge history 😄

PR #278, opened 2021-03-09
refactor: Remove Global call

AndrewCiambrone:
### Summary of Changes

Related to https://github.com/amundsen-io/amundsenmetadatalibrary/pull/270

### Tests

Fixed tests so that they use the new configuration. 

### Documentation

### CheckList

- [ ] PR passes `make test`

### DIFF
--- a/metadata_service/proxy/neptune_proxy.py
+++ b/metadata_service/proxy/neptune_proxy.py
@@ -13,7 +13,7 @@ from amundsen_gremlin.gremlin_model import WellKnownProperties
 from amundsen_gremlin.neptune_bulk_loader.api import (
     NeptuneBulkLoaderApi, get_neptune_graph_traversal_source_factory)
 from amundsen_gremlin.script_translator import ScriptTranslatorTargetNeptune
-from amundsen_gremlin.test_and_development_shard import get_shard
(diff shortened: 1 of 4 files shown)


metadata_service/proxy/neptune_proxy.py
```diff
@@ -65,6 +65,10 @@ def __init__(self, *, host: str, port: Optional[int] = None, user: str = None,
         if port is not None:
             raise NotImplementedError(f'port is not allowed! port={port}')
 
+        if client_kwargs.get('ignore_neptune_shard', False):
+            _reset_for_testing_only()
```
AndrewCiambrone, line 69: @friendtocephalopods I know this is a private method but is this okay to use here?

verdan:
@AndrewCiambrone this is really nice. I like the way it serves only the specific proxy/code which needs this. 💯

metadata_service/proxy/neptune_proxy.py
friendtocephalopods: Hmm we shouldn't need to do do this here. If we'd like some os-level flag to not shard neptune in tests, let's add it to https://github.com/amundsen-io/amundsengremlin/blob/master/amundsen_gremlin/test_and_development_shard.py#L17 and just return None. Then we won't need this stuff propagating to clients of the library.

PR #289, opened 2021-04-07
feat: Create users in metadata backend via API

sewardgw:
### Summary of Changes

- Exposed a `PUT` method for the user details API resource
- Added corresponding swagger docs
- New function to create or update user was added to the base proxy
  - Function was implemented for Neo4j, added as placeholder for other proxies

This implementation heavily relies on the validation and pre-processing being done as part of `UserSchema().load({...})` to validate the inputs of the data provided in the API. As new fields / validation rules are added to the User schema there should _hopefully_ not be any changes to this API. When invoking the API any number of fields can be provided as long as the minimum required fields are provided.

##### Sample invocations

To invoke the API directly, locally:

```python
import requests as r
data = {'email': 'an_email@aol.com', 'first_name': 'first', 'last_name': 'last'}
url = 'http://0.0.0.0:5002/user'
resp = r.put(url, json=data)
resp.json()
# {
#   "profile_url": null,
#   "is_active": true,
#   "slack_id": null,
#   "user_id": null,
#   "manager_fullname": "",
#   "last_name": "last",
#   "full_name": null,
#   "team_name": null,
#   "employee_type": null,
#   "manager_email": null,
#   "email": "some_email@aol.com",
#   "other_key_values": {},
#   "github_username": null,
#   "first_name": "first",
#   "display_name": null,
#   "role_name": null,
#   "manager_id": null
# }
```

Similarly, an invalid payload will return the reason why:

```python
import requests as r
data = {}
url = 'http://0.0.0.0:5002/user'
resp = r.put(url, json=data)
resp.json()
# {'message': 'User inputs provided are not valid: {\'_schema\': [\'"display_name", "full_name", or "email" must be provided\']}'}
```

To invoke from the amundsen frontend:

```python

from amundsen_application.api.utils.request_utils import request_metadata
from amundsen_application.models.user import load_user

def get_create_user() -> User:    
    url = '{0}{1}'.format(app.config['METADATASERVICE_BASE'], '/user')
    _email_ = 'abc_test_email_123'
    user_info = {
        'email': _email_,
        'user_id': _email_,
        'first_name': 'Firstname',
        'last_name': 'Lastname'
    }
    user_resp = request_metadata(url=url, method='PUT', json=user_info)
    return load_user(json.loads(user_resp.json()))
```

### Tests

- Tests added for PUT method as well as Neo4j proxy

### Documentation

- Added corresponding swagger docs under `swagger_doc/user/detail_put.yml`

### CheckList

- [ ] PR passes `make test`

### DIFF
--- a/metadata_service/proxy/neo4j_proxy.py
+++ b/metadata_service/proxy/neo4j_proxy.py
@@ -18,6 +18,7 @@ from amundsen_common.models.table import (Column, ProgrammaticDescription,
                                           Reader, Source, Stat, Table, Tag,
                                           User, Watermark)
 from amundsen_common.models.user import User as UserEntity
+from amundsen_common.models.user import UserSchema
 from beaker.cache import CacheManager
 from beaker.util import parse_cache_config_options
 from flask import current_app, has_app_context
@@ -41,6 +42,12 @@ _CACHE = CacheManager(**parse_cache_config_options({'cache.type': 'memory'}))
 
 # Expire cache every 11 hours + jitter
 _GET_POPULAR_TABLE_CACHE_EXPIRY_SEC = 11 * 60 * 60 + randint(0, 3600)
+
+
+CREATED_EPOCH_MS = 'publisher_created_epoch_ms'
+LAST_UPDATED_EPOCH_MS = 'publisher_last_updated_epoch_ms'
+PUBLISHED_TAG_PROPERTY_NAME = 'published_tag'
+
 
 LOGGER = logging.getLogger(__name__)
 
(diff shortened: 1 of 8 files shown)


dorianj:
LGTM pending a test!

metadata_service/proxy/neo4j_proxy.py
```diff
@@ -958,6 +965,61 @@ def get_user(self, *, id: str) -> Union[UserEntity, None]:
...
+        RETURN usr, usr.%s = timestamp() as created
+        """ % (user_props, CREATED_EPOCH_MS, user_props, CREATED_EPOCH_MS))
+
+        try:
+            tx = self._driver.session().begin_transaction()
+            result = tx.run(create_update_user_query, user_data)
+
+            user_result = result.single()
+            if not user_result:
+                raise RuntimeError('Failed to create user with data %s' % user_data)
```
allisonsuarez, line 993: I don't think this should be a RuntimeError since no result existing would likely be an issue with neo4j, not this program.

sewardgw: I think that's a fair point. This specific error was used to keep in line with the existing paradigm for how Neo4j integration errors are raised, e.g.:

- https://github.com/amundsen-io/amundsenmetadatalibrary/blob/master/metadata_service/proxy/neo4j_proxy.py#L403
- https://github.com/amundsen-io/amundsenmetadatalibrary/blob/master/metadata_service/proxy/neo4j_proxy.py#L499

There are 6-7 places where this is used the the thought was that reusing this exception would make it easier to catch and handle these errors if we only need to watch for one type of exception. 

Given this, if you'd like to continue with a more descriptive exception hat are your thoughts on something akin to a `Neo4jErrorException` class and raise that here in place of the `RuntimeError` the code above (with migration of other runtime errors to this new exception occurring through a refactor in the future)?

tests/unit/proxy/test_neo4j_proxy.py
```diff
@@ -622,6 +622,31 @@ def test_get_user_other_key_values(self) -> None:
...
+            mock_session = MagicMock()
+            mock_driver.return_value.session.return_value = mock_session
+
+            mock_transaction = MagicMock()
+            mock_session.begin_transaction.return_value = mock_transaction
+
+            mock_run = MagicMock()
+            mock_transaction.run = mock_run
+            mock_commit = MagicMock()
+            mock_transaction.commit = mock_commit
```
madison-ookla, line 640: I'm not sure if this is a code consistency thing, but because the `mock_driver` is a mock object, you don't have to set the attributes to other mock objects (since it will create them dynamically). I believe you can do the following:
```suggestion
        with patch.object(GraphDatabase, 'driver') as mock_driver:
            mock_transaction = mock_driver.return_value.session.return_value.begin_transaction.return_value

            mock_run = mock_transaction.run
            mock_commit = mock_transaction.commit

            # Testing the mock below remains the same
```

sewardgw: Indeed, good catch

dorianj (approved):
Thanks @sewardgw for taking this over the line, and thank you @alran for doing this originally! This should be compatible with the API shape that Square is using, so should be able to remove your customization

PR #293, opened 2021-04-19
fix: Sort reports alphabetically

sinkuladis:
Signed-off-by: sinkuladis <sink.vlad@gmail.com>

### Summary of Changes

With this PR table reports are sorted alphabetically.

### Tests

Added test for _get_reports method.

### Documentation

No documentation added or modified.

### CheckList

- [x] PR passes `make test`

### DIFF
--- a/metadata_service/proxy/atlas_proxy.py
+++ b/metadata_service/proxy/atlas_proxy.py
@@ -428,7 +428,7 @@ class AtlasProxy(BaseProxy):
         parsed_reports = app.config['RESOURCE_REPORT_CLIENT'](reports) \
             if app.config['RESOURCE_REPORT_CLIENT'] else reports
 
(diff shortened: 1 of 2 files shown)


dorianj:
This is a behavior change, not a refactor. This seems reasonable, can you explain the user-facing reason that sorting is desired / what this affects?

sinkuladis:
> This is a behavior change, not a refactor. This seems reasonable, can you explain the user-facing reason that sorting is desired / what this affects?

Without sorting, users would see reports in the order they are retrieved from Atlas. I believe this is a random order or order from the most recent to oldest report, I need to double-check this. Sorting would make reports list more consistent, and in the case of a long list, it makes it easier to look for the right one.

mgorsk1 (approved):
lgtm !

PR #299, opened 2021-04-30
Modified RuntimeError to NotFoundException in neo4j_proxy (metadata s…

aparna-cimpress:
Try block in  `_put_resource_description`  & `put_column_description` method throws RuntimeError when table or column doesn’t exist. I have changed it to throw NotFoundException with 404 status code.

### DIFF
--- a/metadata_service/proxy/neo4j_proxy.py
+++ b/metadata_service/proxy/neo4j_proxy.py
@@ -400,7 +400,7 @@ class Neo4jProxy(BaseProxy):
                                                              'key': uri})
 
             if not result.single():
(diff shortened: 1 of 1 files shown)


aparna-cimpress:
Linked with https://github.com/amundsen-io/amundsen/pull/1027 PR.

dorianj:
Hi @aparna-cimpress, thanks for the contribution. This generally LGTM, however, the Amundsen project [moved to a monorepo](https://github.com/amundsen-io/rfcs/pull/31). This sub-repository is now in read-only mode so that users with forks can have a smooth migration path (or if we find an urgent security issue that we want to backport), but new changes all need to go on the new monorepo, which is the `amundsen` repo's main branch

